PR #2, opened 2018-07-16
Skeleton generated from edx/cookiecutter-django-ida

ormsbee:
Everything needs to change -- this is just the baseline starter output of the IDA cookiecutter.


ormsbee:
Per @bradenmacdonald, since this is just the cookiecutter template, it can merge without review.

PR #4, opened 2018-08-08
First version of REST API

symbolist:
**What works:**

- `/api/v1/collections/` with the standard actions for collections.
- `/api/v1/bundles/` with the standard actions for bundles.
- `/api/v1/bundles/<uuid>/files/` lists files in snapshot of the latest version of the bundle. Files can be added or replaced via POST and deleted via DELETE. Each file add, replace or delete creates a new version.
- `/api/v1/bundles_versions/` lists bundle versions.
- `/api/v1/bundles_versions/<bundle_uuid>,<version_num>/files/` lists files in snapshot of the version.
- All the actions above can be performed from the DRF interface available at `/api/v1/`.

**Todo for future rounds:**
- An endpoint to POST a full snapshot as an archive for a bundle.
- Add `expand` field to `/bundles/` and `/bundle_versions/` to inline the files list.
- Add filters to the endpoints.
- A lot more.

**Non-goals:**
- Permissions, authentication, throttling, pagination or anything that will not impact the shape of the API.
- In the future some actions (like deletes) or editing of certain fields may not be possible. But that can be decided once the data model has stabilized.


symbolist:
@pomegranited @bradenmacdonald @ormsbee This is ready for a discussion.

blockstore/apps/bundles/models.py
```diff
@@ -0,0 +1,203 @@
...
+    @receiver(snapshot_created)
+    def listen_for_snapshot_creation(**kwargs):
+        bundle_uuid = kwargs['bundle_uuid']
+        snapshot_digest = kwargs['snapshot_digest']
+
+        bundle = Bundle.objects.get(uuid=bundle_uuid)
+        versions = list(bundle.versions.order_by('-version_num')[:1])
+        next_version_num = versions[0].version_num + 1 if versions else 1
+
+        bundle.versions.create(
```
pomegranited: In final point of the [Blockstore Design: General Themes/Concepts](https://openedx.atlassian.net/wiki/spaces/AC/pages/737149430/Blockstore+Design#Blockstore-Design-genera), it says: 

> versions are meaningful, and not every edit of every file spawns a new version. A version is like a "commit" in that sense.

It seems like creating versions automatically when new snapshots (files) get added to a bundle violates this statement?

ormsbee:
Regarding the URLs:

> /api/v1/collections/ with the standard actions for collection models.
> /api/v1/bundles/ with the standard actions for bundle models.
> /api/v1/bundles/<uuid>/versions/ lists versions of a bundle.
> /api/v1/bundles/<uuid>/versions/<version>/files/ lists files in snapshot of the version.
> /api/v1/bundles/<uuid>/files/ lists files in snapshot of the latest version of the bundle. Files can be added or replaced via POST and deleted via DELETE. Each file add, replace or delete creates a new version.

Could we move bundle_versions to its own resource instead of making it a child of bundles? I could see wanting to do queries on bundle_versions across bundles (e.g. "What bundle versions were most recently created?"). Doing the common "all the versions of this bundle" then becomes a call to that resource with a querystring bundle_uuid={}, and uniquely identifying a BundleVersion is {bundle_uuid}@{version_num}. We could make a few versions special like {bundle_uuid}@latest which redirects to the most recently created version... though maybe that's a bit too magical...

Do we need a separate URL to list files, as opposed to a flag on versions?

blockstore/apps/api/v1/serializers/snapshots.py
```diff
@@ -0,0 +1,49 @@
...
+            return self.context['request'].build_absolute_uri(default_storage.url(data_location))
+
+
+class FileSerializer(serializers.Serializer):
+
+    path = serializers.CharField()
+    public = serializers.BooleanField()
+    size = serializers.IntegerField(read_only=True)
+    url = FileHyperlinkedIdentityField(read_only=True)
+    data = FileDataField()
```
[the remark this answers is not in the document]
ormsbee: We could try deriving it from the filename? What do you think clients will want to do with the information?

blockstore/apps/bundles/models.py
ormsbee: This is my fault for not making any clear sketch of how drafts should work.

I was thinking that snapshots represented a state of the file tree with all file data, like a commit in git. Snapshots and versions would mostly be 1:1 with each other, though there might be rollback situations where multiple versions point to the same snapshot. But snapshots were still meaningful in the sense that they represent a desired state of the Bundle's files and not an intermediate editing state.

For the intermediate editing state, I thought we'd have a directory of resources that was basically a copy-on-write (from the most recent BundleVersion) scratch space. And that when you hit "publish", Blockstore would take the contents of that scratch space and create a new snapshot from it. (Insert hand waving)

blockstore/apps/api/v1/serializers/snapshots.py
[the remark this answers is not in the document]
pomegranited: If the file is to be made available for download or viewing in a browser, then knowing the mimetype is useful so that it can be served with the appropriate Content-Type and/or Content-Disposition headers.  Various cloud storage folks have APIs to store this info in their object storage protocols (e.g [Google Cloud](https://cloud.google.com/storage/docs/metadata), [OpenStack](https://developer.openstack.org/api-ref/object-store/#create-update-or-delete-container-metadata), [AWS S3](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html)).

The mimetype information can be inferred from the filename (unreliably) or better yet, from the file content itself, and so it makes some sense to store/update this when the file data itself is received.  But I don't know if it's better to store this here in the in Blockstore model, or let it be metadata that's set elsewhere.

blockstore/apps/bundles/models.py
pomegranited: > But snapshots were still meaningful in the sense that they represent a desired state of the Bundle's files and not an intermediate editing state.

Ok, I think I understand.. 

A "snapshot" is a deliberate action by the user, saying "I want to save this state", and so these should have annotations attached, and we'll provide a way for users to redo/undo between them.

But a "version" is every single edit that gets made to the bundle, and so it's automatically created, as shown above.

Is that right?

blockstore/apps/api/v1/serializers/snapshots.py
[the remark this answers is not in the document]
symbolist: Hmm. Yeah, it makes sense to have it stored as part of the metadata somewhere. We will need it for cases where for example a web file manager wants to decide which preview component to pass the file URL too. What it be useful to store the modified date or any other metadata that file managers commonly need?

blockstore/apps/bundles/models.py
symbolist: The way I understand it is that a snapshot refers to the state of bundle content while the `Version` model is for keeping track of the order of snapshots of a bundle. What you are referring to by "(tracking) deliberate action by the user" is something that the `Version` model should keep track of. I think the later needs some more thinking. I have some thoughts/notes that I'll be posting later on the data model PR.

symbolist:
@ormsbee

> Could we move bundle_versions to its own resource instead of making it a child of bundles? 

Yeah, that sounds useful.

> We could make a few versions special like {bundle_uuid}@latest which redirects to the most recently created version... though maybe that's a bit too magical...

Since this is additive, we can add it later if it turns out to be useful.

> Do we need a separate URL to list files, as opposed to a flag on versions?

1. If there are bundles with large number of files (hundreds), the list view response is going to become very large. Also large bundles are likely going to have more versions too.
2. As far as I understand we can only get one object from S3 at a time (and other storages may have the similar restrictions) and one will be needed for each snapshot.

How about an `extra_fields` get parameter to inline the file resources when getting the detail of a version (or bundle)?

symbolist:
@pomegranited Sorry if it is not clear but it should not be possible to create a version resource by directly POSTing. Version resources are auto-created when updating files in the bundle. I'll add a check to disable such requests in the view.

blockstore/apps/bundles/models.py
symbolist: @bradenmacdonald One complexity is that unlike in Studio, a bundle can be much larger than a Unit. What happens if two people are editing different parts of a bundle at the same time and one hits publish? 

Some possible options:

1. The API has restrictions on concurrent editing of the same bundle.
2. The API provides information about concurrent edits taking place so that the authoring interface can warn authors.
3. The API has support for multiple draft branches.
4. The API lets author deal with this themselves.

Are there any requirements from product regarding this?

ormsbee: >My preference and the way I had been thinking about it is: that we have numbered versions like 1,2, 3, 4 and a draft version. The numbered versions are always immutable and only created explicitly by a "publish" action that the user takes. The draft version is a mutable version and is the working copy where changes can be made. Hitting publish creates a new numbered version that matches the draft version, and hitting "discard changes" reverts the mutable draft version to the latest numbered version if any. Hitting "revert to version N" produces a new numbered version identical to the specified older version.

I agree with how drafts work conceptually as a working copy, and the publish/discard/revert workflow you outline here. From the user's perspective, the draft can be something that is presented as a version-in-flight. But I think it should be modeled as something entirely separate from BundleVersions, since it differs so much in mutability, data representation, update notifications, associated metadata, discovery, etc.

> @bradenmacdonald One complexity is that unlike in Studio, a bundle can be much larger than a Unit. What happens if two people are editing different parts of a bundle at the same time and one hits publish?
>
>Some possible options:
>
>1. The API has restrictions on concurrent editing of the same bundle.
>2. The API provides information about concurrent edits taking place (including the diff of the draft from the last published version) so that the authoring interface can warn authors.
>3. The API has support for multiple draft branches.
>4. The API lets author deal with this themselves.

I think we leave the door open for (3) in the long term by not making Bundles aware of their drafts -- i.e. have the Drafts push into BundleVersions but don't have Bundles assume they have one and only one Draft. But (2) gets us 90% of the way towards what we need for a lot less complexity. A simple log showing who else is doing editing will let the authors coordinate amongst themselves, and course teams are small enough where I think that should be sufficient most of the time. Teams with really high concurrency needs can still use the github-push workflow that they probably already use.

bradenmacdonald: ^ That all sounds good to me :)  I do think that (2) gives us the most value in the short term.

requirements/base.txt
```diff
@@ -3,5 +3,6 @@ django-extensions
 django-rest-swagger
 django-waffle
 djangorestframework
+git+https://github.com/alanjds/drf-nested-routers.git
 edx-auth-backends
 pytz
```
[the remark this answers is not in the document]
symbolist: This has added in https://github.com/open-craft/blockstore/pull/3.

symbolist:
@bradenmacdonald @pomegranited @ormsbee This is ready for review.

@ormsbee I have moved `/bundle_versions/` to the root of the API. However, I spent a couple of hours trying to get `expand` parameter working for inlining the files in the `Bundle` and `BundleVersion` resources but wasn't able to get it working as nicely as I wanted to. So unless it is critical for anything right now, I would prefer to defer it to the next round of updates to the API when we add support for draft/publish.

ormsbee:
@symbolist: Sounds good, thank you.

requirements/base.txt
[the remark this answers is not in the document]
pomegranited: 👍

blockstore/apps/api/v1/serializers/snapshots.py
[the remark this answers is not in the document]
pomegranited: Ah yep, last modified date, content type, whether it should be served inline or as an attachment -- these are all potentially useful pieces of metadata.

Worth thinking about, but I see they're not relevant to this ticket.  The needs for this data will arise out of subsequent use cases, when we start trying to serve out these bundle files to browsers.  So long as it's OK to expand the bundle data model to suit these, I'm fine to let this thread go for now.

pomegranited:
> > We could make a few versions special like `{bundle_uuid}@latest` which redirects to the most recently created version... though maybe that's a bit too magical...
> Since this is additive, we can add it later if it turns out to be useful.

👍  to this, but I have to say: can we avoid using the word `latest`?  I find it ambiguous..  I encountered this issue when trying to discern how our course lists are ordered in the platform.  Does it mean "oldest", or does it mean "most recent"?  I'd much prefer words like `@newest`, or `@current`, or even `@live`.  My 2¢ 😄

blockstore/apps/api/v1/views/bundles.py
```diff
@@ -0,0 +1,52 @@
...
+
+    lookup_field = 'uuid'
+    lookup_url_kwarg = 'bundle_uuid'
+    lookup_value_regex = UUID4_REGEX
+
+    queryset = Bundle.objects.all()
+    serializer_class = BundleSerializer
+
+
+class BundleVersionViewSet(viewsets.ReadOnlyModelViewSet):
```
pomegranited: Since `BundleVersion.change_description` is mutable, I think we need a `PUT` endpoint too (though the other fields can't be edited.

blockstore/apps/bundles/models.py
pomegranited: ^^ sounds good to me too, and letting humans negotiate to resolve concurrent edits is much better than trying to sort them out automatically.

But despite the fact that we've agreed that:
> The numbered versions are always immutable and only created explicitly by a "publish" action that the user takes. 

A new [`BundleVersion` is still auto-created when a `BundleSnapshot` is created](https://github.com/open-craft/blockstore/blob/master/blockstore/apps/bundles/models.py#L132-L150).  I would say that this should be resolved as part of a subsequent PR, but this REST API is reflecting this same confusion.

Instead, I would expect that:
* the REST API (and/or the models underneath) would create (only) a `BundleSnapshot`  when changes are made to a `Bundle` or its constituent `BundleFiles`.
  That is, neither the REST API nor the models should be auto-creating `BundleVersion` objects.
* the REST API would provide an explicit "Create `BundleVersion`" endpoint, that can be invoked when a human decides to label a particular snapshot with a version.
* the REST API would also provide a "Modify `BundleVersion`" endpoint, which allows users to edit the parts of the `BundleVersion` which are mutable -- currently just the `change_description`, but in the future, maybe metadata.
* the REST API would provide a way to list/view the `BundleSnapshot` list, so that `BundleVersion`s can be created from the desired snapshot.
* We'll figure out how to deal with Drafts later, since they're special.
  They are mutable, or perhaps, applying updates to a Draft will destroy previous the Snapshot.

@symbolist  @ormsbee  Am I missing something crucial here?

blockstore/apps/api/v1/views/bundles.py
```diff
@@ -0,0 +1,52 @@
...
+from ..serializers.bundles import BundleSerializer, BundleVersionSerializer
+
+
+class BundleViewSet(viewsets.ModelViewSet):
+    """
+    ViewSet for Bundle model.
+    """
+
+    lookup_field = 'uuid'
+    lookup_url_kwarg = 'bundle_uuid'
```
pomegranited: Can we also filter `Bundle`s associated with a `Collection`?

Ok to defer this to a subsequent PR, since this is just a first pass at the REST API.

blockstore/apps/api/v1/views/snapshots.py
```diff
@@ -0,0 +1,105 @@
...
+            file = snapshot.files.get(path, None)
+            if not file:
+                raise http.Http404
+
+            file_serializer = FileInfoSerializer(
+                file,
+                context=self.get_serializer_context(),
+            )
+            response_data = file_serializer.data
+        return Response(response_data)
```
pomegranited: Lots of duplicated code between `list` and `retrieve`.. can these be refactored without being confusing?

blockstore/apps/bundles/store.py
```diff
@@ -269,3 +273,20 @@ def create_snapshot(self, bundle_uuid, paths_to_files):
...
+        files = dict(files)
+        files[str(path)] = FileInfo(
+            path=path, public=public, size=data.size, hash_digest=file_hash.digest()
+        )
+        return self._create_snapshot(bundle_uuid, files)
+
+    def snapshot_by_removing_path(self, snapshot, path):
+
+        files = {p:snapshot.files[p] for p in snapshot.files if p != path}
+        return self._create_snapshot(snapshot.bundle_uuid, files)
```
pomegranited: This `snapshot_by_removing/adding` optimisation is very nice 👍

blockstore/apps/api/v1/views/bundles.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class BundleVersionViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    ViewSet for BundleVersion model.
+    """
+
+    lookup_fields = ('bundle__uuid', 'version_num')
+    lookup_url_kwargs = ('bundle_uuid', 'version_num')
+    lookup_value_regexes = (UUID4_REGEX, VERSION_NUM_REGEX)
```
pomegranited: I couldn't figure out how to get a list of all `BundleVersion` objects for a given `Bundle` -- both the `bundle_uuid` and the `version_num` must be in the path, else there's a 404.

This would be useful -- is it possible?

blockstore/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,382 @@
...
+        url = reverse('api:v1:bundleversionfile-list', kwargs={
+            'bundle_uuid': self.bundle_version.bundle.uuid, 'version_num': self.bundle_version.version_num
+        })
+        response = self.client.get(url)
+
+        # response.wsgi_request is the Django Request object and not the DRF Request object.
+        response.wsgi_request.parser_context = {
+            'kwargs': {
+                'bundle_uuid': self.bundle_version.bundle.uuid,
+                'version_num': self.bundle_version.version_num,
```
bradenmacdonald: Nit: When your test cases have to know this muchabout the internals of the serializers etc. (the above few lines as well as things like `'detail_view_name': BundleFileReadOnlyViewSet.detail_view_name`), it might be better to just check the output directly, instead of using the serializer to generate the expected output?

e.g.
```
response = self.client.get(url)
self.assertEqual(
    response.data,
    [
        {
            "path": 'a/file-1.html'
            "public": False,
            "size": 28,
            "url": response.data[0].url,
            "data": response.data[0].data,
        },
        {
            "path": 'b/file-2.txt'
            "public": False,
            "size": 20,
            "url": response.data[1].url,
            "data": response.data[1].data,
        },
    ],
)
```

I also find that much easier to read and understand.

blockstore/apps/bundles/models.py
```diff
@@ -91,16 +91,15 @@ class Bundle(models.Model):
...
-    # collection = models.ForeignKey(
-    #     Collection, related_name="bundles", related_query_name="bundle"
-    # )
+    collection = models.ForeignKey(
+        Collection, related_name="bundles", related_query_name="bundle", editable=False
+    )
 
     slug = models.SlugField(allow_unicode=True)  # For pretty URLs
-    description = models.TextField(max_length=10_000)
+    description = models.TextField(max_length=10000)
```
bradenmacdonald: Nit: why remove the separator here?

blockstore/apps/bundles/store.py
```diff
@@ -269,3 +273,20 @@ def create_snapshot(self, bundle_uuid, paths_to_files):
         logger.info("Created Snapshot %s for Bundle %s", snapshot.hash_digest.hex(), bundle_uuid)
 
         return snapshot
+
+    def snapshot_by_adding_path(self, bundle_uuid, files, path, public, data):
```
bradenmacdonald: Please add docstrings

blockstore/apps/bundles/store.py
```diff
@@ -269,3 +273,20 @@ def create_snapshot(self, bundle_uuid, paths_to_files):
...
+
+        file_hash = FileInfo.generate_hash(data)
+        data_write_location = "{}/data/{}".format(bundle_uuid, file_hash.hexdigest())
+        if not default_storage.exists(data_write_location):
+            default_storage.save(data_write_location, data)
+        files = dict(files)
+        files[str(path)] = FileInfo(
+            path=path, public=public, size=data.size, hash_digest=file_hash.digest()
+        )
+        return self._create_snapshot(bundle_uuid, files)
```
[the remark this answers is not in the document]
symbolist: I have created a method `_save_file(self, bundle_uuid, path, file, public=False)` and moved the common code to it. There is a lot that is going to be added to this file (including draft/publish flow, error checking) so any code refinements can be added later too when this code is refactored.

blockstore/apps/api/v1/views/bundles.py
symbolist: Added todo to LX-53.

blockstore/apps/api/v1/views/bundles.py
symbolist: Yes, we should have all the filters that make sense. I looked at this briefly but there are [different options](http://www.django-rest-framework.org/api-guide/filtering/) for adding filters plus (like in this case) we will also want to add related resources fields to at least `CollectionSerializer` and `BundleSerializer`. Its a reasonably sized chunk of work so have added it to LX-53.

blockstore/apps/api/v1/views/bundles.py
symbolist: `<bundle_uuid>,<version_num>` is a composite identifier. Its like a `usage_key` incorporates the `course_key`, `block_type` and `uuid`. To be able to list `BundleVersions` for a `Bundle`, we will need a filter `/api/v1/bundle_versions/?bundle_uuid=<bundle_uuid>`.

There are some details related to it under "Identifying Resources" in the [edX REST API Conventions](https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/18350757/edX+REST+API+Conventions).

Todo for filters has been added to LX-53.

blockstore/apps/api/v1/tests/test_views.py
symbolist: Hmm. Interesting points.

The reason for adding this is that DRF creates a wrapper object around the Django `Request` object which has all this data but for some reason it is not setting it on the API client response so we cannot use it. One option would be to add it ourselves but I just decided to go with this because of time constraints. Can look at this in the next round.

As for checking the output directly, it depends on whether we want to check all the details that have been tested in the serializer test cases as well? These tests are only running through the happy path right now, so we can update them in the next round when we add more.

blockstore/apps/bundles/models.py
symbolist: Fixed.

blockstore/apps/bundles/store.py
symbolist: Added missing docstrings.

blockstore/apps/api/v1/views/snapshots.py
symbolist: Moved some of the common code to `get_bundle_version_or_404()`.

blockstore/apps/api/v1/tests/test_views.py
symbolist: Added a point about this to LX-53.

blockstore/apps/api/v1/views/snapshots.py
```diff
@@ -0,0 +1,105 @@
...
+from ...constants import FILE_PATH_REGEX
+from ..serializers.snapshots import FileInfoSerializer
+
+
+class BundleFileReadOnlyViewSet(viewsets.ViewSet):
+    """
+    Read only ViewSet for Snapshot files.
+    """
+
+    detail_view_name = 'api:v1:bundleversionfile-detail'
```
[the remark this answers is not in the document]
symbolist: @pomegranited @bradenmacdonald The edX REST API guidelines say:

> Design your API from the perspective of the consumer, NOT the perspective of your underlying implementation.  For example:
> * If the underlying implementation requires accessing multiple models or multiple apps/projects, this does not need to be reflected in a public interface.  From the perspective of the consumer, it's simply one thing they are requesting.
> * Keep CRUD operations together within its corresponding resource.  Why have the client go to one resource to read it and then another resource to write it?

So we can assume that the Snapshot is an internal storage detail and as far as consumers of the API are concerned the files are associated with the Bundle. I have removed references to snapshots from the view docstrings and it is not used anywhere in the URLs. I hope that makes things clearer?

symbolist:
@pomegranited @bradenmacdonald I have addressed your feedback and noted the additive changes needed in LX-53.

blockstore/apps/api/v1/views/snapshots.py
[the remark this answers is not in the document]
bradenmacdonald: That makes sense and sounds good to me.

bradenmacdonald:
Thanks @symbolist. I won't have time for a formal review so don't wait for my approval here :)

blockstore/apps/api/v1/tests/test_views.py
bradenmacdonald: > As for checking the output directly, it depends on whether we want to check all the details that have been tested in the serializer test cases as well?

Personally, I do believe that it's better to err on the side of simple, readable, understandable test cases, even when that means testing multiple things at once (and having redundant tests), compared to test cases with lots of mocking and extra helper code added to ensure that they are only testing one thing at a time, which usually end up being harder to read and understand. But I recognize that both are valid approaches and it's fine to leave this as-is.

blockstore/apps/api/v1/views/bundles.py
pomegranited: 👍

blockstore/apps/api/v1/views/bundles.py
pomegranited: 👍

blockstore/apps/api/v1/views/bundles.py
pomegranited: 👍

blockstore/apps/bundles/store.py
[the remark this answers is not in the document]
pomegranited: Beautiful.  Thank you! 👍

blockstore/apps/api/v1/views/snapshots.py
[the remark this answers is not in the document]
pomegranited: Thank you yes, that helps 👍

blockstore/apps/api/v1/views/snapshots.py
pomegranited: Lovely 👍

blockstore/apps/bundles/models.py
pomegranited: I'm still not clear on why [saving a `BundleSnapshot` automatically triggers a new `BundleVersion` to be created](https://github.com/open-craft/blockstore/blob/master/blockstore/apps/bundles/models.py#L132-L150).  But it seems like I'm the only one who's confused about this, so it feels silly to keep blocking the review on this point.

So am letting this go for now.. though I expect that it'll come up later once we stick a GUI on this 😄

symbolist:
@pomegranited @bradenmacdonald Thanks for your review here!

@pomegranited Yes, correct; the `BundleVersion` creation will be updated when the draft/publish workflow is added.

@ormsbee Would you like to do a review as well or should I go ahead with merging?

blockstore/apps/api/v1/tests/test_views.py
symbolist: Sure, it is possible this makes more sense for these tests. There were already a number of things to setup from scratch in this round so would be happy to think more deeply about it in the next round.

blockstore/apps/api/v1/views/snapshots.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class BundleFileViewSet(BundleFileReadOnlyViewSet):
+    """
+    ViewSet for files in the most recent BundleVersion of a Bundle.
+    """
+
+    detail_view_name = 'api:v1:bundlefile-detail'
+
+    def create(self, request, bundle_uuid, version_num=None):
```
pomegranited: @UmanShahzad I'm having trouble posting multiple files to this endpoint -- is there a way to make this work?  Or as an alternative, can you enable PUT requests to this endpoint which appends files to the bundle?

See [openedx-blockstore-relay#1](https://github.com/open-craft/openedx-blockstore-relay/pull/1/files#diff-e44130bcee7479b4f9a12331518df666R98) for my WIP attempt to post multiple files in a single request.

symbolist: @pomegranited Since each file is considered a resource, right now this will require posting each file independently. I have added a point to LX-53 to investigate how we can add support for posting multiple files at once.

P.S. HTTP 400 Bad Request: The tagged Github id is not a participant of this PR. ;)

UmanShahzad: HTTP 501: I cannot fulfill this request

pomegranited: Sorry for the mis-tag -- I blame tab-complete 😄 

I'm cool with posting multiple file upload requests, but that's not working with the current API.  A POST replaces all the files in a bundle, and PUTs aren't allowed.  Can you allow PUTs here?

symbolist: That is weird. Can you check what requests the DRF web interface is sending? Because I can add multiple files via that. And there is a test for this too: https://github.com/open-craft/blockstore/pull/4/files#diff-3959f0e2e005ddd8ad570fb6febdea94R332. Since the view logic is custom, merely changing the HTTP verb should not change the behavior.

pomegranited: Ah oops..  you're totally right @symbolist .  I thought that POST would replace all the files, but it doesn't, it's just additive.  If I want replace them all, I'll need to iterate over the undesired files and DELETE them.

Thanks for your help!

symbolist:
I have re-organized the commits.

@ormsbee confirmed on chat that I can go ahead with merging.

Here are the test results:
```
bash-4.4# py.test
============================== test session starts ===============================
platform linux -- Python 3.6.5, pytest-3.7.2, py-1.5.4, pluggy-0.7.1
Django settings: blockstore.settings.test (from ini file)
rootdir: /blockstore/app, inifile: pytest.ini
plugins: django-3.4.2, cov-2.5.1
collected 31 items                                                               

blockstore/apps/api/v1/tests/test_serializers.py ....                      [ 12%]
blockstore/apps/api/v1/tests/test_views.py ................                [ 64%]
blockstore/apps/bundles/tests/test_models.py .                             [ 67%]
blockstore/apps/bundles/tests/test_store.py ..                             [ 74%]
blockstore/apps/core/tests/test_context_processors.py .                    [ 77%]
blockstore/apps/core/tests/test_models.py ...                              [ 87%]
blockstore/apps/core/tests/test_views.py ....                              [100%]

=========================== 31 passed in 4.68 seconds ============================
```

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5485](https://openedx.atlassian.net/browse/OSPR-5485) so that we can track it in Jira.

PR #5, opened 2018-08-27
LX-27 Bundle format and conventions.

symbolist:
The documentation is in `docs/bundle_format.rst` and the example bundles in `bundle-examples/`.

**Some Highlights**
- Prefers explicit over implicit/magic.
- Currently its very flexible about how files can be organized. It is not necessary for the rules to be tighter but we can still do that.
- Allows specifying more than one independently linkable XBlock.


docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+The Blockstore does not parse or validate most files itself. However, some conventions must be followed by clients
+to allow interoperability between environments:
+
+* There are no constraints on how the files in a Bundle can be organized.
+* OLX files must use the **.olx** extension. Files with a different extension may be ignored or not treated as
+  OLX by plugins and clients.
+* A bundle.json file must be added by authoring clients which specifies the shareable components, bundle
+  dependencies and assets. See below for details.
+* Every XBlock element in an OLX file must have an **id** attribute. The value of this attribute will be used
+  to construct the **block_key**. In a single Bundle, all ids must be unique. Blockstore will not validate ids however.
```
symbolist: This is what the `url_name` field is currently.

symbolist: Basically decoupling filenames and ids. Though would it make sense to keep them coupled?

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+  "meta": {
+    "version": 1
+  },
+
+
+main_component
+~~~~~~~~~~~~~~~~~~~~~~
+
+The optional **main_component** field can be used to specify the file path to an XBlock which is the primary XBlock
+of the Bundle that authors should use.
```
symbolist: Do we need this? Or can just the `components` field do?

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+The optional **dependencies** field is a dictionary of **bundle_version_alias: bundle_version** pairs. If a file from
+another Bundle is to be referenced, a version of that Bundle must be specified in this field.
+
+.. code-block::
+
+  "dependencies": {
+    "a": 21d45e735e134c41ae3b24fde26d4369@8,
+    "b": b97c9907ecd54f4eb5f4c7eb51dd58e3@2
```
symbolist: Actually thinking now that this should be dict; for future expandability.

symbolist:
@bradenmacdonald @ormsbee Here is my take on this. And its ready for your review.

CC @pomegranited

docs/bundle_format.rst
bradenmacdonald: Checking that IDs in a bundle are unique might be slightly expensive, if it requires scanning the contents of all OLX files in the bundle.

It would be much easier to just say that the runtime will generate the full unique ID for each XBlock from the OLX file name combined with an optional id attribute specified within the OLX. But that would not be backwards compatible with current courses, which allow authors to specify any custom ID slug (`url_name`) for each XBlock.

Although maybe we can collect+index+check IDs during the compilation phase (["Transform 1" in my proposal](https://openedx.atlassian.net/wiki/spaces/AC/pages/803405867/Compositor+Architecture+Proposal)). That would be efficient.

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+* Nested XBlocks in separate OLX files can be referenced by adding a **path** attribute to the element which points
+  to the relevant OLX file. Blockstore does not parse or validate any paths itself. However plugins may do so.
+* XBlock Runtimes should provide a service which can be used to read a file given a **path** or to get a
+  public URL for it (if it is specified in assets field). XBlocks will be responsible for replacing paths
+  in any assets and resources they are using with URLs.
+
+Paths
+-------------------------
+
+Paths are URIs of the form **[//bundle_version_alias]path**.
```
bradenmacdonald: Add: "...where `path` starts with a `/`"  ?

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+
+    "components": {
+      "/video_teaser.olx",
+      "/mcqs/mcq1.olx",
+      "/mcqs/mcq2.olx",
+      "/mcqs/mcq3.olx"
+    }
+
+It is valid to not specify **main_component** and **components**. This may be useful in some cases like having a
+shared Bundle with branding assets only.
```
bradenmacdonald: Should we say explicitly that it's also valid to specify _both_ a `main_component` and a `components`?

docs/bundle_format.rst
bradenmacdonald: A dict makes sense. Also, the example above is not valid JSON since it's not quoted.

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+another Bundle is to be referenced, a version of that Bundle must be specified in this field.
+
+.. code-block::
+
+  "dependencies": {
+    "a": 21d45e735e134c41ae3b24fde26d4369@8,
+    "b": b97c9907ecd54f4eb5f4c7eb51dd58e3@2
+  }
+
+The XBlock Runtime will provide a service which allows reading files from Bundles.
```
bradenmacdonald: I would actually prefer just to have a new field type like `StaticAssetField` which holds an opaque string, and which provides methods to the XBlock like `get_absolute_url()`, `get_source_path_string()` (returns `/static/foo.blah` for assets stored in contentstore as today's platform does, or `//blah/foo` for Blockstore assets).

By not having a dedicated field API for this today, we run into several issues, and I think a `StaticAssetField` would solve them.

Just as we don't want Blockstore to know about XBlocks, I don't think we want XBlocks to have to know about Blockstore, and especially not to use a Blockstore service.

bundle-examples/bundle-sequential/introduction/a6e1.olx
```diff
@@ -0,0 +1 @@
+<html id="a6e14e9cc2e74ec197609d6955be63c8" filename="/introduction/a6e1.html"/>
```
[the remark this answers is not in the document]
ormsbee: So this is probably out of scope for this PR, but I sometimes think it would be nice if a unit could be some kind of non-XML container file format, like how YAML can embed multiple documents in a single file, but more general. Almost like MIME multipart messages, e.g.:

```
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary=xblock
<insert other headers here for vertical info>

--xblock
Content-Type: text/html
XBlock-Type: html5

<html tags we don't have to worry about being valid XML>

--xblock
Content-Type: text/markdown
XBlock-Type: problem

(capa markdown text here)
```

docs/bundle_format.rst
bradenmacdonald: > Perhaps we can make all links relative instead of starting with //? If link convention is relative, then the links should still work if the Bundle is exported into a subdirectory somewhere on a user's machine.

Yeah that makes a lot of sense to me.

docs/bundle_format.rst
symbolist: Thanks for catching that. And will update to a dict.

bundle-examples/bundle-blocks/questions/problem1.olx
```diff
@@ -0,0 +1,10 @@
+<problem id="9c2aacff3e964be39a2bafdf9231587e" display_name="Multiple Choice" markdown="&#10;&gt;&gt;Where does the Blockstore store blocks content?&lt;&lt;&#10;&#10;( ) MySQL&#10;( ) Mongo&#10;( ) S3&#10;">
```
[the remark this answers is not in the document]
symbolist: I have created a new task to track these OLX improvement ideas: LX-54.

bundle-examples/bundle-sequential/introduction/a6e1.html
```diff
@@ -0,0 +1,6 @@
...
+<li><strong>Versioned content is the core of Blockstore, and plugin extensibility is focused around annotating that content.<br /></strong>Things that create, transform, update, or execute content live outside Blockstore. Plugins know when content has been changed, but they don't modify content. Plugins maintain their own data and APIs. Plugin data changes can happen outside of the lifecycle of the content itself. This means that an export of the same version of a Content Bundle will always yield the same authored content, but may yield different plugin metadata (example: new tags that were added). Also, versions are meaningful, and&nbsp;not every edit of every file spawns a new version. A version is like a "commit" in that&nbsp;<span class="inline-comment-marker" data-ref="0840d980-d444-414d-a516-c0dd226579d8">sense.</span></li>
```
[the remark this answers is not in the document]
symbolist: I created the content in Studio and then reorganized the export. It must something that the text editor created.

bundle-examples/bundle-sequential/introduction/a6e1.olx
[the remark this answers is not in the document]
symbolist: @bradenmacdonald Usually the convention is to use `url_name=<id>` to refer to nested blocks. I have replaced that with `path=<path to olx file>`. The `filename` attribute is `html` block specific (and perhaps to blocks which inherit from it) so haven't changed it; though the identifier will need to be the full path.

Adding to LX-54.

bundle-examples/bundle-sequential/introduction/b53d-2.html
```diff
@@ -0,0 +1,2 @@
+<p>This is what the first version of Blockstore looks like.</p>
+<p><img src="//control_center/files/blockstore.jpg" alt="Blockstore V1" width="1103" height="687" /></p>
```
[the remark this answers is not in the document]
symbolist: Yup CC0.

https://pixabay.com/en/toy-box-toys-play-duplo-1916163/

bundle-examples/bundle-sequential/sequence1.olx
```diff
@@ -0,0 +1,15 @@
+<sequential display_name="What is Blockstore?">
+
+  <vertical id="4d5ec9e8e6cb4f5c861ac266b3bc7c13" display_name="Introduction">
+    <html path="/introduction/d68c.olx"/>
+    <html path="/introduction/b53d.olx"/>
+  </vertical>
+
+  <vertical path="//control-center/info1.olx"/>
```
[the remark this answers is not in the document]
symbolist: Oh yes, that is correct! I re-did the paths at the end so must have missed this.

docs/bundle_format.rst
symbolist: Hmm. Yeah, this is a bit tricky. To be clear my since Blockstore is not expected to parse the files so it will not be verifying them. Anything inside the OLX file will have to be verified at some other step.

By file name do you mean just the filename or the whole path because if the former then we will need to make sure it is unique across all directories. And whichever it is, changing it implies deleting the block and creating a new one. Right?

docs/bundle_format.rst
symbolist: Yup; will do.

docs/bundle_format.rst
symbolist: Ah, good idea! But I think we will need something on the Runtime as well which has the config, right?

bundle-examples/bundle-sequential/sequence1.olx
```diff
@@ -0,0 +1,15 @@
+<sequential display_name="What is Blockstore?">
+
+  <vertical id="4d5ec9e8e6cb4f5c861ac266b3bc7c13" display_name="Introduction">
```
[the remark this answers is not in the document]
symbolist: Noted in LX-54 as a block specific OLX improvement. Though this may be tricky because the element type is used to get the XBlock class.

docs/bundle_format.rst
symbolist: I do see the proposal uses this convention. I thought about this quite a bit and landed on this variation with the dependency name as the domain name for the following reasons:

* We will have to reserve a root dir path for `links` and both at the Blockstore and client level add handling for "You cannot create a directory with this path" error messages.
* It applies to a subset of authoring situations since its not clear how the export-bundle-dependencies-into-subdirectory will work if the editing has to happen across multiple bundles. A more generic solution would be to have all bundles exported at the root of a single directory. And then, lets say, the XBlock workbench can load the correct files just like other LMS/Studio environments would. Maybe something like a directory backed Blockstore. Also note that even in course exports right now, there are no "correct paths".
* This solution also allows us to nicely support dependencies across Blockstores. For example, the author only needs to download the bundle they want to edit locally and for any dependencies not available locally the workbench can be configured to cache bundles from a remote Blockstore.

Would like your thoughts on the above!

symbolist:
@bradenmacdonald 

> The one thing that I think is still missing here is a way to specify the bundle type (the bundle holds a whole course, the bundle holds a sequential, the bundle holds a collection of problems, the bundle holds a collection of video files but no XBlocks, etc.)

> e.g. "type": "olx/unit" in the bundle.json file?

We can definitely add metadata for that. Can you clarify the use-cases for it as well? Is there other metadata we would want? Should we change the `main_component` value into a dict so that this can be in place? What is specified if there is no `main_component` or even `components`?

symbolist:
@bradenmacdonald @ormsbee I think I have responded to all the comments/questions in this pass. I am going to make the fixes and re-organize the points a bit tomorrow into sections.

docs/bundle_format.rst
ormsbee: I think it's fine to have just the `components` field.

docs/bundle_format.rst
bradenmacdonald: > By file name do you mean just the filename or the whole path

I guess I mean the whole path, slugified. And yeah, if a block ID changes, it's effectives the same as deleting it and creating a new one as far as the LMS is concerned.

docs/bundle_format.rst
bradenmacdonald: Yes, the runtime would have to know how to get the absolute URL for a given blockstore static asset path.

bradenmacdonald:
@symbolist 

> Can you clarify the use-cases for it as well?

There will be many types of Bundles in blockstore serving different purposes, and we need some way of knowing the author's intent for a bundle. Hence, the "type".

For example, a bundle may be a collection of images, including both the source .ai files and the resulting .svg or .png image files.

Another bundle type may represent a course, including an `outline.json` and several `unit-blah.olx` files.

Another bundle type might represent a content library.

> Is there other metadata we would want? Should we change the main_component value into a dict so that this can be in place? What is specified if there is no main_component or even components?

Well, for a bundle that is a collection of images (or other static assets), it wouldn't make sense to have a `components` field at all. We can probably just assume that every file in the bundle is a useful asset that could be used elsewhere. Though for that sort of bundle we may want a list of `"assets": [...]` which distinguishes the usable final assets (`.svg`, `.png`, or `.pdf` files) from the source files (`.ai`, `.psd`, `.docx`, etc.)

bundle-examples/bundle-sequential/introduction/a6e1.olx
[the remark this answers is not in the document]
ormsbee: Can we use XInclude for the file includes? Having Blockstore understand how to follow and piece together an HTML block based on various attributes sounds like it's getting into XBlock runtime territory, but XInclude is a standard spec that lxml supports.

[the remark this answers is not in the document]
pomegranited: I [tried to parse the HTML content to append it to the OLX node](https://github.com/open-craft/openedx-blockstore-relay/pull/1/commits/b18ab614a286bdb841adff84b1556304414a8f4b#diff-e44130bcee7479b4f9a12331518df666R68), and found that many of the HTML blocks in the DemoX course at least don't parse as valid XML.

I would expect that XInclude-ing that content would have the same issue?

docs/bundle_format.rst
symbolist: @ormsbee Great! This seems a bit arbitrary to me too!

@pomegranited This field is for assisting with discovery and not for the runtimes. For example, we will have a Library which lists all the bundles which are shared with the authors. One of the requirements is that authors should be able to specify the granularity of sharing, i.e. they should be able to say: "This vertical is meant to be used as a whole" or "This sequential is meant to be used as a whole". So the blocks specified in `components` are what the authors will see and be able to select for inclusion in their course and pathway.

I am going to remove the `main_component` field here. So in the data transfer command you were working on, only the root block id that was specified to the command should be listed in the `components` field.

docs/bundle_format.rst
symbolist: @ormsbee Did you look at this id generation idea? Just want to make sure we are all on the same page.

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+will have lower priority than the XBlock specified in **main_component**.
+
+.. code-block::
+
+    "components": {
+      "/video_teaser.olx",
+      "/mcqs/mcq1.olx",
+      "/mcqs/mcq2.olx",
+      "/mcqs/mcq3.olx"
+    }
```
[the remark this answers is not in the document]
symbolist: Ah, the perils of hand-editing json!

bundle-examples/bundle-sequential/introduction/a6e1.olx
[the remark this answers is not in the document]
symbolist: @ormsbee Since the content files will be stored in separate storage and Blockstore will not be reading them (just saving and tracking URLs), I do not see where Blockstore will get the chance to piece together the OLX?

Having said so I see it makes sense to move to a standard attribute for the embedded content for the `html` block. Should we add using `xinclude` in such scenarios as a recommendation for other blocks to adopt as well?

@pomegranited Yes, that is not going to work most of the time. `xinclude` will not have the same issue since we can use [`parse=text`](https://www.w3.org/TR/xinclude/#include_element). For now I would recommend saving the html in a separate file as is done in course exports.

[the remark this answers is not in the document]
ormsbee: I suppose it depends on how much we want Blockstore itself to be able to piece things together. When doing rendering for preview or for the student view, we're probably going to want the entire Unit most of the time. If stitching together the resources that are supposed to be shown inline (so things like HTML modules, not subtitles) requires XBlock runtime understanding, then either:

1. The XBlock runtime is coupled to Blockstore and runs in-proc -- which we explicitly want to avoid.
2. It's going to have to make multiple network roundtrips to get the pieces it needs to stitch together.
3. It's all taken care of in a compile phase that sits with an XBlock runtime somewhere else.

But I wonder if we can get make Blockstore just smart enough to pull these things inline without requiring an XBlock runtime. So we use XInclude with `parse=text` to pull in things like HTML and later Markdown and Python files in Capa problems. Then a client is able to ask "Give me this Unit file, but processed with the includes." Blockstore can understand and cache that, and basically serve up whole all the OLX content of a Unit in one resource without actually needing to understand OLX. And since pretty much everything is immutable, we can aggressively cache it at the Blockstore layer.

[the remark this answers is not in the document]
symbolist: Hmm I see. I do not see any technical limitations with doing this since XInclude is a standard with library support. And the current "inclusion" mechanism actually is a bit weird because for nested XBlocks, the block element is present both in the parent OLX and the OLX of the block. So if we can move to something more standard that sounds like a plus to me.

symbolist:
@bradenmacdonald @ormsbee I have improved and reorganized the docs. There are two new files `docs/bundles.rst` and `docs/olx.rst`. For now I have left the existing file in because there are open threads. So you can go ahead with another round of review.

@bradenmacdonald 

> We can probably just assume that every file in the bundle is a useful asset that could be used elsewhere.

Yes, it would be hard to restrict linking to individual files manually if someone has access to the Bundle because all files are going to be used one way or another by the components of that Block anyway and are going to be available in the learner context.

> Well, for a bundle that is a collection of images (or other static assets), it wouldn't make sense to have a components field at all.

Yes, this is why the `components` field is optional.

> Though for that sort of bundle we may want a list of "assets": [...] which distinguishes the usable final assets (.svg, .png, or .pdf files) from the source files (.ai, .psd, .docx, etc.)

Basically similar to what we are doing by specifying some blocks in `components`? That is, only a subset of them which would be useful for other bundles to link to should be listed in Bundle previews. I like the idea! Unfortunately, I am running into one of the two hardest problems in CS here. * sigh *

> There are only two hard things in Computer Science: cache invalidation and naming things. -- Phil Karlton

Can you suggest a name for the field? I think authors should be independently able to specify files-other-than-olx which should be listed for linking in Bundle previews and files which should be accessible via internet accessible URLs since these two may be non-overlapping sets.

> There will be many types of Bundles in blockstore serving different purposes, and we need some way of knowing the author's intent for a bundle. Hence, the "type".

Hmm. Can you suggest something for this? I'll just add it in. Not sure I have sufficient context to be able to present the optimal option.

bundle-examples/bundle-sequential/sequence1.olx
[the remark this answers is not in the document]
symbolist: Fixed.

bundle-examples/bundle-sequential/introduction/b53d-2.html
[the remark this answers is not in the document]
symbolist: Fixed.

bundle-examples/bundle-sequential/introduction/a6e1.html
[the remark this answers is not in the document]
symbolist: Removed.

docs/bundles.rst
```diff
@@ -0,0 +1,126 @@
...
+meta
+~~~~~~~~~~~~~~~~~~~~~~
+
+The **meta** field is required and specifies the document format version.
+
+.. code-block::
+
+  "meta": {
+    "version": 1
+  },
```
symbolist: @ormsbee Since we want the file to be author-friendly, can we collapse this dict to just `"version": 1`?

docs/olx.rst
```diff
@@ -0,0 +1,8 @@
+OLX Conventions
+===================
+
+* Nested XBlocks in separate OLX files can be referenced by adding a **path** attribute to the element which points
+  to the relevant OLX file. The path can be to an OLX file in the same Bundle or in another Bundle.
+* The path to an OLX file may be used as a component of the XBlock ID. Therefore moving an OLX file or changing its
+  name will be equivalent to deleting the XBlock and creating a new one. A block element in an OLX file may
+  also have an **id** attribute. The value of this attribute will be used when constructing the **block_key**.
```
symbolist: @bradenmacdonald If you have a better explanation for this, let me know and I'll edit it in.

docs/bundles.rst
ormsbee: Yeah, it's just that I wanted some way to disambiguate the version of the bundle vs. the version of the file format. Totally open to something else here.

bundle-examples/bundle-sequential/introduction/a6e1.olx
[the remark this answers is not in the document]
bradenmacdonald: I think that combining files as text using Xincludes or option 3 (prepare+optimize+combine files during a compilation phase) make sense.

docs/olx.rst
bradenmacdonald: @symbolist This is fine but it's not very clear what will happen since it says "the path..._may_ be used". It would be better if we can decide for sure how things will be used. Although I'm personally still not sure of the best approach, so this may be a WIP that we update as we try things out.

docs/bundle_format.rst
pomegranited: 👍 thanks for the explanation @symbolist , I'll change the code on my data transfer PR to suit.

docs/bundle_format.rst
```diff
@@ -0,0 +1,106 @@
...
+================
+
+The Blockstore does not parse or validate most files itself. However, some conventions must be followed by clients
+to allow interoperability between environments:
+
+* There are no constraints on how the files in a Bundle can be organized.
+* OLX files must use the **.olx** extension. Files with a different extension may be ignored or not treated as
+  OLX by plugins and clients.
+* A bundle.json file must be added by authoring clients which specifies the shareable components, bundle
+  dependencies and assets. See below for details.
```
pomegranited: Could you check my logic below to see if I understand the role of the `bundle.json` file for the initial work to transfer Open edX content into Blockstore?

When transferring an Unit (vertical) from Open edX to a Blockstore Bundle:
* The Unit's OLX should be the only path listed in the `components` list.
   Other OLX and related files from the Unit's XBlocks might be transferred to the Bundle too, but they will not be listed in the `components` list.
* If any of the Unit's XBlocks reference static assets, then those static assets should also be copied over to the Bundle.
  But the static assets should not be listed in the `assets` list, because they (on their own) are not "sharable" with other Bundles.
  (These static assets live in the Bundle, and so if other Units are transferred to other Bundles, then separate copies of these asset files will be stored.)
* The `dependencies` list will be empty, because the Open edX XBlock doesn't know anything about Blockstore, and so can't have dependencies on other Bundle files yet.

When transferring a Content Library from Open edX into a Blockstore Bundle:
* Each XBlock in the Library will have an OLX path in the `components` list.
* If an XBlock references static assets, then those static assets should also be copied over to the Bundle.
  But the static assets should not be included in the `assets` list, because they (on their own) are not "sharable".
* If multiple XBlocks in a Content Library reference the same asset file, then that asset file is only stored once in the Bundle.
* The `dependencies` list will be empty.

And we might have some other way to transfer just static assets from an Open edX `course` into a Blockstore Bundle:
* Each static asset will be put in the `assets` list, so that they can be shared with other Bundles.
* The `components` and `dependencies` lists would be empty.

bundle-examples/bundle-sequential/introduction/a6e1.olx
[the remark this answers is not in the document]
pomegranited: With the data transfer PR, I have the option to nest the XBlock OLX directly inside the `vertical` OLX, essentially performing the compilation prior to transfer.

Or I can keep them separate, and use `XInclude` to include them.

Which mechanism is preferred?

Note: this is a separate question from how the `.html` content should be included in the HTML OLX.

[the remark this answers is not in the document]
symbolist: @pomegranited and I discussed this over chat. My recommendation was that for now, until there has been a decision on the OLX changes, sticking with the conventions used in course export sounds like the straightforward path to take.

docs/bundle_format.rst
symbolist: Yup, that is mostly accurate! I'll see if I can clarify the language in the doc even more.
 
> But the static assets should not be listed in the assets list, because they (on their own) are not "sharable" with other Bundles.

For now the purpose of the `assets` field is to list files that should be accessible by a client with a public URL. For example, if we have an html block with an image, the `.html` is a non-olx file but it does not need to be accessible from a cdn but the image file must be so that it can be fetched by browsers.

Right now, there is no mechanism for specifying non-olx files that should be shown to other authors in Bundle previews as shareable items. Am thinking about this right now.

docs/bundles.rst
symbolist: Okay, great. I am going to simplify this to:

```
"schema": 0.1
```

Later we can also add a schema for this file and have it validated on POST with something like https://github.com/Julian/jsonschema.

bradenmacdonald:
@symbolist 

> > There will be many types of Bundles in blockstore serving different purposes, and we need some way of knowing the author's intent for a bundle. Hence, the "type".
> 
> Hmm. Can you suggest something for this? I'll just add it in. Not sure I have sufficient context to be able to present the optimal option.

I think we just need each bundle.json to have a `"type"` field, and for now we can define just a few possible values: `olx/unit`, `olx/collection`, `olx/course`, and perhaps `static` ? We just need a rough idea for now; we can always refine the possible values and their meaning later

symbolist:
@bradenmacdonald @ormsbee Okay, so Schema 0.1 is ready for a final review.

@bradenmacdonald I have added the `type` field as you suggested. I have also removed the OLX related points that are still under discussion for now.

We may need to enhance the fields related to static assets too at some point but what is exactly needed will become clearer once we look at the compilation-to-LMS-format phase.

docs/bundle_format.rst
pomegranited: @symbolist  Ah cool -- I wasn't clear on what the `public` field was intended for either.  For now, I'll use `public=True` for the `asset` files, and set `public=False` for the `.html` related files, but will leave both of these out of the sharable `assets` list in `bundle.json`.

Should `.olx` files be `public` too?

bundle-examples/bundle-sequential/introduction/a6e1.olx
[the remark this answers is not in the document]
pomegranited: @ormsbee 
>  I sometimes think it would be nice if a unit could be some kind of non-XML container file format, like how YAML can embed multiple documents in a single file, but more general. Almost like MIME multipart messages

If we want to to ditch OLX, then why not use the  the JSON data (as YAML) as exported by the Course Blocks API?  We're already relying on it for native apps, and so will need to ensure that the same data can be exposed for Blockstore bundles.

I'm having a whinge, because I'm currently trying to wrench the transcript assets out of the video OLX.  The xmodule [inexplicably removes the nice `/static` path prefixes](https://github.com/edx/edx-platform/blob/b85d6f031c1d6590a184a873fcadb3b2d42024bd/common/lib/xmodule/xmodule/video_module/transcripts_utils.py#L383-L384), making it impossible to scrape out these paths using the [`static_replace`](https://github.com/edx/edx-platform/blob/b85d6f031c1d6590a184a873fcadb3b2d42024bd/common/djangoapps/static_replace/__init__.py#L151) mechanism that works for other xblocks.  But if I could have used `student_view_data` instead of OLX, I'd already have everything I need in structured form, and wouldn't need to scrape anything:
```json
{
   "student_view_data":{
      "duration":null,
      "all_sources":[
         "https://s3.amazonaws.com/edx-course-videos/mit-6002x/6002-Tutorial-00010_100.mov"
      ],
      "transcripts":{
         "en":"http://localhost:8000/courses/course-v1:edX+DemoX+Demo_Course/xblock/block-v1:edX+DemoX+Demo_Course+type@video+block@636541acbae448d98ab484b028c9a7f6/handler_noauth/transcript/download?lang=en"
      },
      "only_on_web":false,
      "encoded_videos":{
         "fallback":{
            "url":"https://s3.amazonaws.com/edx-course-videos/mit-6002x/6002-Tutorial-00010_100.mov",
            "file_size":0
         },
         "youtube":{
            "url":"https://www.youtube.com/watch?v=o2pLltkrhGM",
            "file_size":0
         }
      }
   }
}
```

docs/bundle_format.rst
symbolist: @pomegranited Being listed in `assets` is the replacement for setting `public=True`. Though this is likely going to evolve overtime as well (for example, we do not have an equivalent for asset locking as it happens in the Files & Uploads section). For now whatever makes the API happy with the least amount of work would be sufficient.

docs/bundles.rst
```diff
@@ -0,0 +1,137 @@
...
+
+.. code-block::
+
+  /bundle.json
+  /mcqs/mcq1.olx
+  /b/c/icon.png
+  /d/README
+  /e/f/
+
+For files in other Bundles, the URI must be a network-path reference which begins with two slash characters followed
```
bradenmacdonald: One problem with this is that authors sometimes use `//example.com/images/image.png` in their content, where `//` gets resolved by the browser to be either `http` or `https`. So we might occasionally get conflicts.

Same issue applies to URLs like `/images/blah.png` - is that intending to refer to a path on the Open edX server, or to the "images" folder in the root of the current bundle?

I'm almost thinking we should use a prefix like `bundle:/image.png`. There's a lot to be said for having no ambiguity.

Also, did we consider @ormsbee's suggestion of "mounting" external bundles on virtual paths within the bundle? I think that was kind of nice. (e.g. `bundle:/deps/otherbundle/image.png`)

docs/bundles.rst
```diff
@@ -0,0 +1,137 @@
...
+two fields: **bundle_uuid** and **version_num**.
+
+If a file from another Bundle is to be referenced, a version of that Bundle must be specified in this field.
+URIs to directories or files in other Bundles use the alias as the authority.
+
+.. code-block::
+
+  "dependencies": {
+    "problems": {
+      "bundle_uuid": "21d45e735e134c41ae3b24fde26d4369",
```
bradenmacdonald: UUIDs are normally written with hyphens - can we add them in here, and should we specify whether or not hyphens are allowed?

docs/bundles.rst
symbolist: I am fully onboard with having the scheme specified in the URI. Saving a few characters for convenience but adding complexity+ambiguity seems like the less desirable option to me.

> Also, did we consider @ormsbee's suggestion of "mounting" external bundles on virtual paths within the bundle? I think that was kind of nice. (e.g. bundle:/deps/otherbundle/image.png)

There was a conversation above: https://github.com/open-craft/blockstore/pull/5#discussion_r213142268. And this is already how the runtimes will have to work. The only difference from your example is that instead of `bundle:/debs/<bundle_alias>/<path_to_file>`, its going to be `bundle://<bundle_alias>/<path_to_file>/` where the `bundle_alias` is the URI authority/domain. Given that XBlocks do not deal with any filesystems (and so there are no libs which can benefit from virtual mounting), I do not see what advantages it provides compared to the points I have mentioned in the conversation thread above. Are there any I am missing?

docs/bundles.rst
symbolist: Ah, yes! The canonical representation uses dashes according to section 3 of https://tools.ietf.org/html/rfc4122#section-4.1. I think I just copied a block id from an olx file here. Will fix this and add the format requirement.

docs/bundles.rst
bradenmacdonald: Ok, makes sense :) I don't think you're missing anything.

bradenmacdonald:
@ormsbee Can we get a final review here? Anything else you think we should change in this initial draft? (We can continue to evolve this as we build out the blockstore and test it in practice.)

symbolist:
@ormsbee Thanks!

I have updated the file URI format to include the scheme as discussed above. Going to merge now.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5484](https://openedx.atlassian.net/browse/OSPR-5484) so that we can track it in Jira.

PR #7, opened 2018-09-28
Initial tagstore implementation

bradenmacdonald:
Work in progress


bradenmacdonald:
@ormsbee @symbolist Could you let me know your thoughts on the direction here? I'm anxious to get some kind of tagging in place ASAP so we can hit our deadlines. CC @nasthagiri too :)

bradenmacdonald:
BTW @ormsbee @symbolist when working on this, I found that the current blockstore code doesn't work with MySQL at all - some of the models define things like a BinaryField with an index that the Django ORM doesn't support on MySQL, and some of the migrations get rejected by MySQL. I've created LX-76 to fix that next sprint.

ormsbee:
Haven't looked through all the code yet, but quick comments on overall approach:

* I'm in favor of the overall direction.
* I'm +1 for dropping asyncio and the Neo4J backend if we won't need it for a while.
* I'm unclear as to what the performance issues around tagging would be.
* Would UUIDs really help with import/export? I mean, sure we'd preserve some kind of ID, but aren't taxonomies meaningfully identified by some set of standard namespace + their own conventions internally?

ormsbee:
And sorry about the BinaryField index thing. I've been running it locally on sqlite, and it never occurred to me that it wouldn't support indexing. :(

tagstore/backends/tagstore_django/models.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class Taxonomy(models.Model):
+    """
+    A taxonomy is a collection of tags, some of which may be organized into
+    a hierarchy.
+    """
+    id = models.BigAutoField(primary_key=True)
+    name = models.CharField(max_length=MAX_CHAR_FIELD_LENGTH)
+    owner_id = models.BigIntegerField(null=False)
```
taranjeet: @bradenmacdonald : Default value for null is `False`. Is there any specific reason why are we adding this?

tagstore/backends/tagstore_django/models.py
```diff
@@ -0,0 +1,78 @@
...
+    class Meta:
+        db_table = 'tagstore_taxonomy'
+
+
+class Tag(models.Model):
+    """
+    A tag within a taxonomy
+    """
+    id = models.BigAutoField(primary_key=True)
+    taxonomy = models.ForeignKey(Taxonomy, null=False)
```
taranjeet: same as above. `null` by default is `False`

symbolist:
> Does not implement "private tags" (user A applies tag T to entity E, but only user A sees that tag). I'm not sure if we want private tags or private taxonomies. One use case for private tags could be users tagging blocks as "favorite", but I can't think of any other use cases.

> Determine if we want to support private tags or private taxonomies

I suspect that this will add a lot of complexity and performance overhead to queries since searching for a tag will mean "Show me items which have has been publicly tagged A or tagged A by me but not tagged A by other users who want to keep the association private". Also it will mean having an identifier for every user in Blockstore.

> The intended path forward would be that this lives in blockstore for now and is only used to tag XBlocks and collections, with taxonomies like "subject area" and "learning outcome".

Is it going to support tagging other non-XBlock entities in Bundles like courses or pathways?

> Add a case-sensitive index to Tag.tag, in addition to the case sensitive one (slight oversight)

This may be sufficient for languages in a particular character-set but is this fully i18n capable? Also how is i18n going to work in general? If I search for "trigonometry" in some other language, will I see results for "trigonometry"? Or will it require adding separate tags to each entity for each supported language? If a tag will be translated to the end-users language before it is displayed to them, in which part of the system will that happen?

> Would UUIDs really help with import/export? I mean, sure we'd preserve some kind of ID, but aren't taxonomies meaningfully identified by some set of standard namespace + their own conventions internally?

This may be needed if we want to store the tag translations in tagstore?

> Allows tags to optionally be hierarchical - tags can exist in a tree with parent-child relationships, e.g. all dogs are mammals, all mammals are animals. This is designed to support learning outcome hierarchies in particular.

> Allows rich searching for entities by tags. e.g. "Find all large animals" will return an entity that was tagged with "large" and "mammal", since it knows that the "mammal" tag is a type of "animal" tag.

> Does not allow other types of relationships between tags other than organizing them into a hierarchy (no support for arbitrary relationships like "dog is similar to wolf"; such advanced graph relationships - which enable other types of taxonomies and fuzzy searches - could be added later but would mean we can't use SQL backends)

The implied constraints/properties of the above seem to be because the expectation is that the queries will be against SQL backends, right? Given that there are going to be a number of search facets, many of which can't/shouldn't be represented as tags, and which will get combined in searches (and also with permissions data), do we not expect the browsing interface queries to go though ElasticSearch? If the Django models have to only store the tags, then we can support a lot more capabilities.

bradenmacdonald:
> Would UUIDs really help with import/export? I mean, sure we'd preserve some kind of ID, but aren't taxonomies meaningfully identified by some set of standard namespace + their own conventions internally?

I guess this depends on our import/export story.

If import/export is done simply by "zipping up" all the files in a bundle and sending it over a network to some other blockstore, then tags wouldn't be preserved at all, without some sort of plugin that help with that. So let's say there's some plugin that writes out the taxonomy+tag data into the tarball for the blockstore bundle on export. But in that case, presumably there needs to be some way for the importing blockstore to know when a taxonomy needs to be imported from the tarball or when the taxonomy already exists in the destination blockstore/tagstore. I was thinking UUIDs could help with that, but it's true that some other sort of standard taxonomy identifer could do the same thing. I just don't know if that exists.

If cross-instance sharing is done by connecting an LMS to a remote blockstore, then it doesn't matter as much. But it would be good if for example, the LMS could understand that a certain taxonomy in the remote blockstore (like a "Open edX system tags" taxonomy) was the same or not.

Another option for detecting and deduping tags on import would be to just hash the taxonomy, because I'm assuming that the taxonomies that most matter for deduplication will also be the sort of taxonomies that rarely change (like learning outcomes from some official source, which can be a static taxonomy for years).

tagstore/backends/django.py
```diff
@@ -0,0 +1,155 @@
...
+            # Convert the set of tags to a set of materialized paths:
+            tags_filter = Q()
+            for tag in tags:
+                tags_filter = tags_filter | (Q(taxonomy_id=tag.taxonomy_uid) & Q(tag=tag.tag))
+            paths = TagModel.objects.filter(tags_filter).values_list('path', flat=True)
+            for path in paths:
+                entities = entities.filter(tags__path__startswith=path)
+        else:
+            for tag in tags:
+                entities = entities.filter(tags__taxonomy_id=tag.taxonomy_uid, tags__tag=tag.tag)
```
pomegranited: If the tag provided in the `TagSet` doesn't exist in the database, then I get a `TypeError` below, because no filters are applied here.  E.g.
```python
dogtag = Tag(1, 'dog')
list(self.tagstore.get_entities_tagged_with_all({dogtag}))
# (0.000) SELECT "tagstore_tag"."path" FROM "tagstore_tag" WHERE ("tagstore_tag"."taxonomy_id" = 1 AND "tagstore_tag"."tag" = 'dog') ORDER BY "tagstore_tag"."tag" ASC; args=(1, 'dog')
# Traceback (most recent call last):
#   File "<console>", line 1, in <module>
#   File "blockstore/tagstore/backends/django.py", line 154, in get_entities_tagged_with_all
#     for e in entities:
# TypeError: 'Manager' object is not iterable
```

tagstore/models/user.py
```diff
@@ -0,0 +1,7 @@
+"""
+User-related types
+"""
+
+from typing import NewType
+
+UserId = NewType('UserId', int)
```
pomegranited: Bugs me that the `UserId` is an integer..  but I guess we can change this if/when we need to?

Makefile
```diff
@@ -60,7 +60,7 @@ html_coverage: ## Generate HTML coverage report
 	${VENV_BIN}/coverage html
 
 quality: ## Run quality checks
-	${VENV_BIN}/pycodestyle --config=pycodestyle blockstore *.py
-	${VENV_BIN}/pylint --rcfile=pylintrc blockstore *.py
+	${VENV_BIN}/pycodestyle --config=pycodestyle blockstore tagstore *.py
+	${VENV_BIN}/pylint --rcfile=pylintrc blockstore tagstore *.py
```
pomegranited: If you also update `.coveragerc`, you can see the coverage missing from the `tagstore` code.  
E.g, if you use this:
```diff
diff --git a/.coveragerc b/.coveragerc
index 4c4c847..5f76dff 100644
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,9 @@
 [run]
 branch = True
 data_file = .coverage
-source=blockstore
+source=
+    blockstore
+    tagstore
 omit =
     blockstore/settings*
     blockstore/conf*
@@ -10,3 +12,7 @@ omit =
     *admin.py
     *static*
     *templates*
+[report]
+exclude_lines =
+    raise NotImplementedError()
+    yield None
```

then the coverage report for this change is:
```
tagstore/__init__.py (100%)
tagstore/backends/django.py (95.2%): Missing lines 36-37,105,126
tagstore/backends/tagstore_django/models.py (100%)
tagstore/backends/tests.py (100%)
tagstore/models/__init__.py (100%)
tagstore/models/entity.py (100%)
tagstore/models/tag.py (100%)
tagstore/models/taxonomy.py (100%)
tagstore/models/user.py (100%)
tagstore/tagstore.py (85.0%): Missing lines 53,57,91-94
-------------
Total:   372 lines
Missing: 10 lines
Coverage: 97%
```

tagstore/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+Convenience module to make importing Tagstore easier.
+"""
+from .tagstore import Tagstore
```
pomegranited: Would like to see a README in this new `tagstore` dir too.

tagstore/models/user.py
bradenmacdonald: What type would you suggest instead? String?

pomegranited: Yep string -- since users in edx-platform are identified by the username, not the numeric ID.

tagstore/backends/django.py
bradenmacdonald: Thanks, fixed in 67c54b2efba10a0d53f9fba957eac3e1e64f1fcd by changing the first line to `entities = EntityModel.objects.all()`

I didn't both writing a test case for this, since manually constructing `Tag` objects that don't exist in the DB is not something the API is supposed to support.

tagstore/backends/django.py
```diff
@@ -0,0 +1,155 @@
...
+        entities = EntityModel.objects
+
+        if include_child_tags:
+            # Convert the set of tags to a set of materialized paths:
+            tags_filter = Q()
+            for tag in tags:
+                tags_filter = tags_filter | (Q(taxonomy_id=tag.taxonomy_uid) & Q(tag=tag.tag))
+            paths = TagModel.objects.filter(tags_filter).values_list('path', flat=True)
+            for path in paths:
+                entities = entities.filter(tags__path__startswith=path)
```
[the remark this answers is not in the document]
bradenmacdonald: Thanks :-)

tagstore/backends/django.py
```diff
@@ -0,0 +1,155 @@
...
+            yield Tag(taxonomy_uid=uid, tag=tag.tag)
+
+    # Tagging Entities ##########################
+
+    def add_tag_to(self, tag: Tag, *entity_ids: EntityId) -> None:
+        """
+        Add the specified tag to the specified entity/entities.
+
+        Will be a no-op if the tag is already applied or does not exist
+        in the given taxonomy.
```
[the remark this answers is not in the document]
bradenmacdonald: Fixed in 67c54b2efba10a0d53f9fba957eac3e1e64f1fcd - I just removed the comment. Not sure why I even wrote that in the first place.

tagstore/backends/django.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def get_taxonomy(self, uid: int) -> TaxonomyMetadata:
+        try:
+            tax = TaxonomyModel.objects.get(pk=uid)
+        except TaxonomyModel.DoesNotExist:
+            return None
+        return TaxonomyMetadata(uid=tax.id, name=tax.name, owner_id=tax.owner_id)
+
+    def _add_tag_to_taxonomy(self, taxonomy_uid: int, tag: str, parent_tag: Optional[str] = None) -> None:
+        if parent_tag:
```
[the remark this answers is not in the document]
bradenmacdonald: True. Fixed in 67c54b2efba10a0d53f9fba957eac3e1e64f1fcd. See the new `test_add_tag_to_taxonomy_idempotent_parent` test case.

Makefile
bradenmacdonald: Thanks, updated.

tagstore/__init__.py
bradenmacdonald: Sure, `README` added in 1463db8ef1cb9b41d1a64c082782f8119a1e0eb9

tagstore/models/user.py
bradenmacdonald: Check out a304535f296bfa73f1cf787fa45f9f3377bbcfa4 in which I used the `EntityId` models for users instead of an integer ID. I think that's more consistent with the design, and I'd always intended for the future possibility of tagging users by representing them with an `EntityId`.

pomegranited: Oo that's lovely.. so even people are Entities!  I can see use cases for that, and lovely that even if they never come up in reality, it's no less or more work to maintain.

Great solution!

tagstore/backends/tagstore_django/models.py
```diff
@@ -0,0 +1,106 @@
...
+    # The tag string, like "good problem".
+    # The migration creates an uppercase index on this field for
+    # case-insensitive searches
```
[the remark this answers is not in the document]
bradenmacdonald: Whoops, I definitely didn't mean to delete that index. Will re-add.

tagstore/tagstore.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def list_tags_in_taxonomy_containing(self, uid: int, text: str) -> Iterator[Tag]:
+        """
+        Get a (flattened) list of all tags in the given taxonomy that contain the given string
+        """
+        # Subclasses can optionally override this method to provide a more efficient implementation.
+        text = text.lower()
+        for tag in self.list_tags_in_taxonomy(uid):
+            if tag.tag.lower().find(text) != -1:
+                yield tag
```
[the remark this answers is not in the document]
bradenmacdonald: Sure, good point.

tagstore/tagstore.py
```diff
@@ -0,0 +1,145 @@
...
+        parent_tag_str: Optional[str] = None
+        if parent_tag is not None:
+            if parent_tag.taxonomy_uid != taxonomy_uid:
+                raise ValueError("A tag cannot have a parent from another taxonomy")
+            parent_tag_str = parent_tag.tag
+
+        self._add_tag_to_taxonomy(taxonomy_uid=taxonomy_uid, tag=tag, parent_tag=parent_tag_str)
+        return Tag(taxonomy_uid=taxonomy_uid, tag=tag)
+
+    def _add_tag_to_taxonomy(self, taxonomy_uid: int, tag: str, parent_tag: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
bradenmacdonald: There is no way to distinguish between `private` and `protected` in Python; a leading double-underscore is ambiguous as it simply means "non-public", but I believe this is fine.

What PEP8 _does_ [say](https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance) is that "If your class is intended to be subclassed, and you have attributes that you do not want subclasses to use, consider naming them with double leading underscores and no trailing underscores."

What I'm trying to say here is external users of this API should not call this method, but subclasses are free to override it. I'll put that in the docstrings explicitly.

tagstore/models/tag.py
```diff
@@ -0,0 +1,12 @@
...
+"""
+from typing import NamedTuple
+
+
+class Tag(NamedTuple):
+    """
+    A tag that can be applied to content.
+    """
+    taxonomy_uid: int  # The Unique ID of the taxonomy which owns this tag
+    tag: str  # The text of this tag, which also serves as its identifier. Unique within this taxonomy.
```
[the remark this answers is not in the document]
bradenmacdonald: Good point. The `tag.tag` thing was bugging me. I'll make that change in a separate PR for easier review.

bradenmacdonald:
I have semi-squashed and rebased this in preparation for merge, but found that it was broken when trying to run with MySQL, which led me down a long rabbit hole of unicode issues. I'm going to flag the fixes for that in a separate PR, and get it reviewed and merged into this before I merge this one.

pomegranited:
@bradenmacdonald  Huh.. I rebased your branch on the latest master when I did [my second round of testing](https://github.com/open-craft/blockstore/pull/7#pullrequestreview-171852303), and the manual checks worked fine.. admittedly I didn't re-run the automated tests though.

Feel free to ping me for review on the 2nd PR whenever it's ready.

bradenmacdonald:
@pomegranited It's ready :) https://github.com/open-craft/blockstore/pull/14

pomegranited:
@bradenmacdonald  Sorry for the delayed feedback, but I'm using the `tagstore.models` interface to build the React frontend for facet filtering, and noticed that [`tagstore.models.Tag`](https://github.com/open-craft/blockstore/blob/4e0cb0d20926f4d6d144171505d256ab3c88fa50/tagstore/models/tag.py#L7) doesn't have a `path` variable.  Should it?  Or do we expect that different backends may present the hierarchy differently?

tagstore/models/taxonomy.py
```diff
@@ -0,0 +1,16 @@
+"""
+A taxonomy is a collection of tags that can be applied to content.
+"""
+from typing import NamedTuple, Optional, NewType
+from .user import UserId
+
+TaxonomyId = NewType('TaxonomyId', int)
```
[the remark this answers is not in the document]
symbolist: Ah makes sense!

tagstore/models/tag.py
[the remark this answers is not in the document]
symbolist: 👍

tagstore/models/taxonomy.py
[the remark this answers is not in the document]
symbolist: Ah makes sense!

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5482](https://openedx.atlassian.net/browse/OSPR-5482) so that we can track it in Jira.

PR #9, opened 2018-10-19
WIP SE-342 LX-76 Make Blockstore work with MySQL

earaujoassis:
A setup to run tests in MySQL instead of SQLite

## Description

A setup to run tests in MySQL instead of SQLite. It defines new dependencies in `requirements.txt` files, creates a new container for the `docker-compose` file, and creates new settings (environment variables) to setup a (MySQL) database instance.

## Testing Checklist
- [ ] Include unit tests
- [ ] All unit tests passes
- [ ] Manually test right-to-left languages and i18n
  of the changes

## Reviewers
- [ ] @UmanShahzad 

### Areas to Consider
- [ ] Database migrations
    - Are they backwards compatible?
    - When they run on production, how long will they take? Will they lock the table?

### DIFF
--- a/blockstore/settings/test.py
+++ b/blockstore/settings/test.py
@@ -3,18 +3,18 @@ import os
 from blockstore.settings.base import *
 
 
-# IN-MEMORY TEST DATABASE
+# MYSQL TEST DATABASE
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': ':memory:',
-        'USER': '',
-        'PASSWORD': '',
-        'HOST': '',
-        'PORT': '',
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'blockstore_test',
+        'USER': environ.get('BLOCKSTORE_MYSQL_USER', 'blockstore_user'),
+        'PASSWORD': environ.get('BLOCKSTORE_MYSQL_PASSWORD', 'blockstore_password'),
+        'HOST': environ.get('BLOCKSTORE_MYSQL_HOST', 'datastore'),
+        'PORT': int(environ.get('BLOCKSTORE_MYSQL_PORT', 3306)),
     },
 }
-# END IN-MEMORY TEST DATABASE
+# END MYSQL TEST DATABASE
 
 # So we don't leave mix Bundle files from test runs with our local dev server.
 DEFAULT_FILE_STORAGE = 'inmemorystorage.InMemoryStorage'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,22 @@ services:
       - "18250:18250"
     volumes:
       - ./:/blockstore/app/:cached
+    depends_on:
+      - "datastore"
+
+  datastore:
+    container_name: datastore
+    image: mysql:8.0
+    ports:
+        - "3306:3306"
+    volumes:
+      - ./datastore_volume:/var/lib/mysql
+    environment:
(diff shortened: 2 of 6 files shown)


bradenmacdonald:
Reviewer here is actually @symbolist  ;)

docker-compose.yml
```diff
@@ -18,6 +18,22 @@ services:
       - "18250:18250"
     volumes:
       - ./:/blockstore/app/:cached
+    depends_on:
+      - "datastore"
+
+  datastore:
+    container_name: datastore
```
bradenmacdonald: For local development, it's probably better to just tell people to run the edX devstack MySQL, which we can use for edx-platform development and blockstore development: https://github.com/edx/devstack/blob/88d22083877005548b31977da99a8092ba1d43b3/docker-compose.yml#L64-L74

docker-compose.yml
```diff
@@ -18,6 +18,22 @@ services:
...
+    image: mysql:8.0
```
bradenmacdonald: We'll be running on AWS RDS or Aurora, which only supports MySQL 5.x. So we shouldn't develop using MySQL 8.

docker-compose.yml
earaujoassis: Good to know, @bradenmacdonald ! Thanks

docker-compose.yml
earaujoassis: I'll update that accordingly. Thank you!

blockstore/settings/test.py
```diff
@@ -3,18 +3,18 @@
...
-        'USER': '',
-        'PASSWORD': '',
-        'HOST': '',
-        'PORT': '',
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'blockstore_test',
+        'USER': environ.get('BLOCKSTORE_MYSQL_USER', 'blockstore_user'),
+        'PASSWORD': environ.get('BLOCKSTORE_MYSQL_PASSWORD', 'blockstore_password'),
+        'HOST': environ.get('BLOCKSTORE_MYSQL_HOST', 'datastore'),
+        'PORT': int(environ.get('BLOCKSTORE_MYSQL_PORT', 3306)),
```
UmanShahzad: You can do `environ.int` I think instead of casting to an int manually.

requirements/local.txt
```diff
@@ -3,3 +3,4 @@
 -r docs.txt
 
 django-debug-toolbar
+mysqlclient
```
UmanShahzad: Installing `test.txt` already installs `base.txt` which has the `mysqlclient` package from the diff above this file.

docker-compose.yml
```diff
@@ -18,6 +18,22 @@ services:
...
+    ports:
+        - "3306:3306"
+    volumes:
+      - ./datastore_volume:/var/lib/mysql
```
UmanShahzad: Do we plan on making the volume map to a local file? I think the syntax you're looking for is:

```yml
    ...
    volumes:
      - datastore_volume:/var/lib/mysql
    ...

networks:
  devstack:
    external:
      name: devstack_default
volumes:
  datastore_volume:
```

symbolist:
Closing in favor of https://github.com/open-craft/blockstore/pull/12.

edx-webhook:
Although this pull request is already closed, I've created [OSPR-5480](https://openedx.atlassian.net/browse/OSPR-5480) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #10, opened 2018-10-22
Drafts

ormsbee:
@bradenmacdonald, @symbolist, @pomegranited: This isn't review-ready (I don't have the REST front-end -- wanted to get some opinions there), but I thought I should post it because it changed a decent amount in the last week from what I thought it'd be. Also I'm not sure, but it seems like the diff here is incorrectly showing some of the stuff @symbolist already merged to master? I dunno, I've messed around with with the store.py a decent amount at this point. I'm sorry if it causes any rebase trouble -- I tried to keep the basic interface the same for existing code, though there are a lot of renames here.

So at a high level:

0. The basic data model of creating Drafts from snapshots, updating drafts, and making new Snapshots from them works.
1. I started with Drafts as something pretty divorced from Snapshots, then conflated them both into the `BundleDataStore` with sort of parallel methods, then split them off into separate `SnapshotStore` and `DraftStore`, with the a one way dependency (`DraftStore` depends on `SnapshotStore`).
2. There was an interlude where I tried to make them share a common base class, but for now they're more like cousin classes with some basic similarities but a lot of differences. Not looking to unify them any more. They're superficially the same in some ways (CRUD operations, file operations, URLs), but they're also pretty different in terms of lifecycle, mutability, and storage.
3. Drafts are copy-on-write as originally discussed.
4. Error checking needs to be much more robust.
5. Another thing I went back and forth on was whether to have a summary file for Drafts or to make the Drafts folder itself be everything. I ended up using the summary file. It's not the best for supporting concurrent editing, but it makes the process of returning the metadata about all the files in a Draft fast and predictable.
6. I'm looking for input on the REST API because there are two very different options that I'm trying to sort in my head.

In terms of the REST API, I started building something parallel to what we have for bundle files, but I was concerned about how the XBlock client would use it. I feel like having a way to do GET/PUT/PATCH with multiple files sent through base64 encoded could be more natural. It gives us some amount of transactionality from the point of view of the client (even if it's not real in the S3 persistence sense, it is real in the "I'm not going to 200 on three files and 400 on the last because I specified an illegal filename" sense). It's also going to be useful for the typical OLX use case, which in our case is going to be a bunch of relatively small XML files (one per Unit) -- though we'd likely have to give it parameters so that it doesn't expand everything, just the file types we're comfortable with. Would appreciate thoughts on this.

Thank you.

I'll update a lot of tickets in the morning.


blockstore/apps/bundles/store.py
```diff
@@ -279,27 +319,184 @@ def _create_snapshot(self, bundle_uuid, files):
...
+            base_snapshot=existing_draft.base_snapshot,
+            files_to_overwrite={**existing_draft.files_to_overwrite, **new_files_written},
+            created_at=existing_draft.created_at,
+            updated_at=updated_at or datetime.now(timezone.utc)
+        )
+        draft_summary_path = self._summary_path(bundle_uuid, name)
+        draft_summary_json = json.dumps(new_draft, cls=BundleDataJSONEncoder, indent=2, sort_keys=True)
+        if default_storage.exists(draft_summary_path):
+            default_storage.delete(draft_summary_path)
+        default_storage.save(draft_summary_path, ContentFile(draft_summary_json))
```
bradenmacdonald: This uses `delete` followed immediately by `save`. I think S3 uses an eventual consistency model for deletes, and if so, is it possible the `save()` might still see the deleted file and then ["modify the filename as necessary to get a unique name"](https://docs.djangoproject.com/en/2.1/ref/files/storage/#django.core.files.storage.Storage.save) ?

blockstore/apps/bundles/tests/test_store.py
```diff
@@ -93,3 +89,57 @@ def test_multiple_snapshots(self, snapshot_created_mock):
...
+                'test.rst': ContentFile(b"This is draft RST content")
+            }
+        )
+
+        # Make sure the second draft persisted properly.
+        self.assertEqual(second_draft, draft_store.get(bundle_uuid, draft_name))
+
+        # Create a new Snapshot from the Draft
+        with draft_store.file_mapping(second_draft) as draft_files:
+            second_snapshot = snapshot_store.create(draft.bundle_uuid, draft_files)
```
bradenmacdonald: Is there some way this could be `second_snapshot = draft_store.commit(draft_name)` or similar, pushing this logic into `DraftStore`?

bradenmacdonald:
@ormsbee Nice!

>  a one way dependency (DraftStore depends on SnapshotStore).

Sounds reasonable.

> whether to have a summary file for Drafts or to make the Drafts folder itself be everything

I posted a minor potential issue with overwriting summary files using S3. I don't suppose we'd also want to consider storing the summary data in MySQL?

> concurrent editing

Can you explain a bit more about how this works with this implementation of drafts? i.e. if two users edit the same bundle, does each user have their own Draft, forked from the same original Snapshot? Or is there only one Draft which is shared by all users? I'm guessing the latter, as I see "TODO: Error if already exists?" in the code.

How are you intending for draft `name`s to be used? I see you've put `studio_draft` as an example name.

bradenmacdonald:
Re the REST API:

> It's also going to be useful for the typical OLX use case, which in our case is going to be a bunch of relatively small XML files (one per Unit)

I think there will be two use cases:

1. Bulk importing a library/course/etc. from Modulestore into Blockstore: yes that would produce tons of small file edits.

2. Editing units in Studio: typically, based on where my head's at in https://github.com/edx/edx-platform/pull/19136 , I think your average set of edits in Studio is going to result in only one `.olx` file changed, because users edit one Unit at a time, and so after they make some changes on a Unit page, it's going to serialize that Unit as OLX and push the updated .olx file back into blockstore's active Draft. The .olx file may be accompanied by one or more `.html` or `.srt` files or similar, but that won't be very common nor will they usually be too numerous I think.

It would definitely be helpful if there is a way to make each push of related files atomic.

pomegranited:
@ormsbee CC @symbolist 
>  I feel like having a way to do GET/PUT/PATCH with multiple files sent through base64 encoded could be more natural. 

Yes!  I want this, on non-draft Bundles too.

To elaborate on @bradenmacdonald 's [first use case above](https://github.com/open-craft/blockstore/pull/10#issuecomment-431934007):

> 1. Bulk importing a library/course/etc. from Modulestore into Blockstore: yes that would produce tons of small file edits.

Yes, this is what happens, and I think we should fix it.

Currently, when you use the [openedx-blockstore-relay `transfer_to_blockstore`](https://github.com/open-craft/openedx-blockstore-relay/blob/master/openedx_blockstore_relay/transfer_data.py#L24) command to copy a block and its children from Open edX to Blockstore, it [`POST`s each file individually to the bundle](https://github.com/open-craft/openedx-blockstore-relay/blob/master/openedx_blockstore_relay/transfer_data.py#L249-L275).  This is because we currently only have one `POST` endpoint for a bundlefile, which [creates a new BundleFile](https://github.com/open-craft/blockstore/blob/master/blockstore/apps/api/v1/views/snapshots.py#L87-L107), appending it to the Bundle.  As a consequence, blockstore [creates a new snapshot/version](https://github.com/open-craft/blockstore/blob/master/blockstore/apps/api/v1/views/snapshots.py#L98-L99) for each little file added to the block, which I think makes no sense?

* We don't want a partial block update to partially succeed, 
* We don't need to snapshot incomplete block updates, and
* To remove old, unneeded bundlefiles, we have to iterate over each one and delete them, which also [creates intermediate and incomplete snapshots](https://github.com/open-craft/blockstore/blob/master/blockstore/apps/api/v1/views/snapshots.py#L113).

To follow REST conventions, I think we should support these methods:
* `POST`: creates new BundleFile(s), appending them to the Bundle and generating a new snapshot (current behaviour).
* `PUT`: atomically replaces the full list of the bundle's BundleFile(s), and generates a new snapshot on success.
  This would be used by `transfer_to_blockstore` to replace an existing Bundle with a full Open edX Block and its children.
* `PATCH`: updates existing BundleFile(s).  This supports the normal [Studio use case #2 noted above](https://github.com/open-craft/blockstore/pull/10#issuecomment-431934007).

blockstore/apps/bundles/store.py
ormsbee: So that used to be an issue with US-East in particular, but even they've upgraded so that read-after-write consistency shouldn't be a problem. You're right that the reason I have the delete/save silliness is that `save` won't overwrite by default. I tried using basic `open` functionality but Django's storages will only open something in binary mode (it seems to ignore binary/string mode parameter and always open in binary), which causes issues with `json.dump`. Anyhow, I'll fix it so it's one atomic operation.

blockstore/apps/bundles/tests/test_store.py
ormsbee: Yeah, you're right. That's a better interface for it, since `DraftStore` already has a reference to the backing `SnapshotStore`.

ormsbee:
Okay, trying to synthesize all the feedback (thank you!)...

Drafts are a staging area for creating new Snapshots, implemented with sorta copy-on-write semantics (since we store the back reference to a base snapshot and only store the delta). The Draft `name` is vague because I thought it would give us more wiggle room for introducing concurrency in the future -- it could be determined at the client level, so that if we wanted to have one shared draft for an application (e.g. Studio), we could. But then we could possibly do something more sophisticated later with per-user Drafts, at the client's discretion.

So since we want:

* atomic writes
* support for many/large files
* preferably one obvious way to do it
* no meaningless, intermediate Snapshots
* drafts to get cleaned up

Maybe Drafts are the only way you create BundleVersions, and your app (whatever it is) creates its own private draft when it wants to, with its own namespace. For instance, a Draft name could be `openedx.studio. transfer_to_blockstore@{uuid}`.

So then the Drafts endpoint has:

* POST to create based off the latest BundleVersion's Snapshot. Files can be included here.
* PATCH to update specific files, which is probably what we'd use the most often
* PUT to do a complete replacement
* DELETE to destroy the draft
* POST for a `commit` command of some sort that would create the Snapshot and also delete the Draft.

We could also throw in a `commit=True` type of parameter so that it would create a new version in one  request (and clean up the resulting Draft).

I think that with the current metadata gathering, we have to process the files in proc at some point (to get the hash data), so I don't know if it's worth supporting straight-to-S3 uploads. I do think that inlining base64 could be useful, particularly on the read side -- say a query param that inlined based on file type.

symbolist:
Sounds like a plan to me! 👍

bradenmacdonald:
Sounds good @ormsbee ! My only question is, if we support named drafts, what if two different drafts are made from the same snapshot, then the first is merged, but now the second one conflicts? Is it just last write wins? (We wouldn't want conflicts that involve `bundle.json` to be resolved via last write wins...)

bradenmacdonald:
Thinking about that a bit more, we might actually want to make blockstore manage `bundle.json` instead of the clients? After all, if the schema changes, it could get messy to have different clients writing different schema versions. Or encoding dates in different formats, etc.

symbolist:
> Thinking about that a bit more, we might actually want to make blockstore manage bundle.json instead of the clients? After all, if the schema changes, it could get messy to have different clients writing different schema versions. Or encoding dates in different formats, etc.

Hmm, this is worth considering. I can think of some other advantages too:
1. Clients do not have to keep track of all the metadata for the whole bundle or worry about what are valid transforms. (Who knows how complex this file will become in the future!)
2. Actions in the UI can be mapped to simple REST operations.
3. Error messages + error handling would be much simpler. Instead of having to deal with "Here are 10 errors in the bundle.json you submitted" responses, it would be much simpler for clients to deal with separate errors for individual operations.

ormsbee:
> Sounds good @ormsbee ! My only question is, if we support named drafts, what if two different drafts are made from the same snapshot, then the first is merged, but now the second one conflicts? Is it just last write wins? (We wouldn't want conflicts that involve bundle.json to be resolved via last write wins...)

Last write wins, or an optional error prompt that comes up when you commit a Draft and the base snapshot is not the current one. The `bundle.json` file only has a few high level pieces, so hopefully it won't be too bad? In any case, I think this is an area where we start with the dumbest thing (last one wins), and make it more sophisticated as we run into real world issues. If Studio uses a single draft across users, it probably won't be a big issue most of the time. If it uses per-user drafts, there's a whole other set of concerns we also have to think about.

>> Thinking about that a bit more, we might actually want to make blockstore manage bundle.json instead of the clients? After all, if the schema changes, it could get messy to have different clients writing different schema versions. Or encoding dates in different formats, etc.

I actually rather like the idea of a file that encodes OLX specific metadata information. I'm okay with having some of that data live in Blockstore in a more first-class way (like "type", "version"), but I'm leery of pushing too much OLX knowledge into Blockstore.

> Hmm, this is worth considering. I can think of some other advantages for the client side too:
> 
> 1. Clients do not have to keep track of all the metadata for the whole bundle or worry about what are valid transforms. (Who knows how complex this file will become in the future!)
>
> 2. Actions in the UI can be mapped to simple REST operations.

Can you expand on this one? I don't think I'm understanding it properly.

> 3. Error messages + error handling would be much simpler. Instead of having to deal with "Here are 10 errors in the bundle.json you submitted" responses, it would be much simpler for clients to deal with separate errors for individual operations.

So this touches more on the overall OLX validation story of Blockstore, which I'm going to consider out of scope for this PR but it's also something that I've had a few ideas on while trying to think through drafts, and I wanted to write them down. Some assumptions:

1. We want _some_ level of OLX validation at the Blockstore service.
2. We want said validation to be an extensible plugin by bundle type, and not built into the core of Blockstore.

We've definitely had [performance pain](https://engineering.edx.org/xblock-lessons-plugin-performance-and-grading-2f85a1d6fb2a) around pluggable interfaces in the past. This may not be as relevant to Blockstore if the validation can be done very quickly, but would be very relevant if validation ever happens out of proc (because it requires an XBlock runtime at some point).

Another factor in this is that at some point down the road, creating a new BundleVersion will trigger notifications that allow Links to be updated, search indexing to happen, etc.

So I was originally thinking that we either had to synchronously block on BundleVersion creation or put some kind of extra metadata around BundleVersions to mark that they were validated. But then it occurred to me that the way things are set up now, we have a separation between Snapshot creation and BundleVersion creation, currently connected by signals. Maybe that's a useful place to fit in asynchronous validation.

Now maybe that's more trouble than it's worth, since it complicates how the client would respond to errors and probably means that we'd have to elevate the Snapshot to be an API concept instead of an internal implementation detail. Not sure if it's a good idea honestly, but it's one that I wanted to throw out there while we're on the topic.

bradenmacdonald:
> I actually rather like the idea of a file that encodes OLX specific metadata information. I'm okay with having some of that data live in Blockstore in a more first-class way (like "type", "version"), but I'm leery of pushing too much OLX knowledge into Blockstore.

Yeah, I think the information that's in `bundle.json` is not too OLX-specific and shouldn't require parsing/understanding OLX.

bradenmacdonald:
@symbolist 

> Lets say there is a file which is specified in bundle.json components field. And the client does a patch request and deletes that file. Should the client submit an updated bundle.json in the same request which has that file removed from components? Or is the validation going to be performed at the time the client tries to do a commit? If Blockstore was managing the bundle.json file, then if a file is removed from the bundle then it can auto remove it from the components list. Or is it okay for the components field to have entries which do not correspond to any files in the bundle?

I think it would be best if some OLX plugin synchronously updated `bundle.json` to remove the component from the `components` list automatically when a client commits a delete of a component's OLX file. Otherwise we have to distribute code about maintaining bundle.json files among every client, and I worry that some clients will have bugs or implementation differences, resulting in broken bundle.json files.

> Some component will have to extract that info, either from the video olx or video files and push it into the search index. Whether that component is written against a pluggable interface or not, is that part of "Blockstore"?

I think that the ElasticSearch store and the APIs required to update should be part of Blockstore, but code that actually populates the ElasticSearch database is not.

The simplest model would be a plugin with an XBlock runtime that can load XBlocks in blockstore then call their [`index_dictionary` method](https://github.com/open-craft/unit-xblock/blob/2097ccb91650ea39383832a8faf768da1b947ed4/unit_xblock/unit_xblock.py#L78-L96) and put that result into ElasticSearch. It still requires an XBlock runtime, but it could be a process that runs in Blockstore, in a Studio celery worker, or as a completely independent process.

For now, it's probably easiest to make that part of Studio, but it would be best to get it running independently (if we can get an independent XBlock runtime).

> Which metadata do we want to encode in the directory structure/file names and which in manifest files? Do we want to strongly prefer one or the other for some reason?

Manifest files:
* Manifest files are more flexible and allow for easier schema changes in the future
* Slower to access (but can be cached so this is probably not a big deal)
* Are more complex and require more author awareness/understanding (per @symbolist's example, authors must ensure that a "Make publicly available" checkbox is set appropriately for every file in the course/bundle's static files area)

Path conventions:
* Fast and simple
* Potentially simpler for authors (forced to explicitly choose when uploading if the file is public or not)
* Harder to change public setting for existing content (have to rename file, which may break things)

So it's probably better to go with manifest files in general?

> Which metadata is the store component + in process plugins + search service going to be aware of? And if there is some, why shouldn't it validate all of that on bundle update?

I think it should be able to read+write all the metadata? And yeah, it makes sense to have plugin(s) validate that data on bundle update.

ormsbee:
@symbolist: Good point about `bundle.json` getting out of sync. @bradenmacdonald: In the choice between manifest files and conventions, I think I'd lean towards convention because I feel like it'll be less likely to yield us illegal states. I guess I'm also wondering if we can get rid of `bundle.json` entirely.

I'm going to make a bunch of assumptions up front so please let me know if I'm off base:

* Hidden-by-design (e.g. .olx files) is determined by the XBlock runtime and developer conventions.
* Hidden-by-policy (e.g. private PDFs) is determined by the team running the course, not by the content author creating the initial sequence -- i.e. if it does live in Blockstore, it lives at the layer where we're composing for a course, alongside other policy like due dates.
* `bundle.json`'s current `schema (version)` and `type` fields could be moved to be attributes of `BundleVersion`.

The components portion of `bundle.json` is a use case that I'm not sure we really need. If we already need some kind of simple OLX analysis plugin to give us the usage key -> file mapping, then it's going to cover the "what are all the XBlocks here" use case without having to mark it explicitly in a manifest. Maybe it would be nice to explicitly label multiple sharable pieces, but I don't know if that's worth designing for at this point -- I think it's enough to say that each Bundle is designed to be reused at the top level, and that more granular reuse is supported but not pretty (i.e. make the common thing easy and the out-of-band thing possible if awkward).

So how about:

* a common entry-point file name based on type/schema (e.g. Bundles of type `olx/sequence` all use `sequence.xml` and `olx/course_run` all use `course_run.xml`). When they're used in Links, the Bundles themselves are going to have more descriptive names, so the local file link will look more like `hawaii_vacation_tips/sequence.olx` which seems adequately descriptive.
* a `/static` dir for all assets that could plausibly be sent directly to the student's browser, like videos and images, so that part is done purely by convention. But Blockstore doesn't care.

So something like:

`hawaii_vacation_tips` = Bundle Name, and this Bundle is a sequence
```
      /sequence.olx = Sequence file with a bunch
      /units/stay_away_from_waikiki.olx
      /units/...
      /static/units/stay_away_from_waikiki/abc.jpg
      /static/units/
```

We don't attempt to associate assets at a level more granular than the Bundle. Maybe that limits possible reuse without refactoring, and that's an okay tradeoff to make to simplify the common path.

bradenmacdonald:
@ormsbee  @symbolist 

> Hidden-by-design (e.g. .olx files) is determined by the XBlock runtime and developer conventions.
> Hidden-by-policy (e.g. private PDFs) is determined by the team running the course,

Makes a lot of sense :)

> The components portion of bundle.json is a use case that I'm not sure we really need.
> I think it's enough to say that each Bundle is designed to be reused at the top level, and that more granular reuse is supported but not pretty

So if I've hand-authored a bundle with a bunch of unit/sequence OLX files that each XInclude their constituent block OLX files (so the bundle has `unit1.olx` which includes `unit1-intro.olx`, `unit1-video.olx`, then the bundle has `unit2`, etc.): in that situation we wouldn't have a mechanism for saying "unit1.olx and unit2.olx are the things designed for reuse elsewhere, not the other OLX files". Right? Yeah, maybe that's fine.

> a common entry-point file name based on type/schema

I like that, but what about bundles that contain many "peer" units for reuse instead of a single entry point file?

> We don't attempt to associate assets at a level more granular than the Bundle.

That's something I'd still really love to have, though, if there's any way to make it work.

ormsbee:
Okay, writing down some thrashing (some recycled from our [previous conversations](https://openedx.atlassian.net/wiki/spaces/AC/pages/793870498/Bundle+Parts)):

Right now, Bundle Versions represent both the grouping of things we change/version by as well as the addressable thing for reuse. These are often but not always going to be the same. When they are not the same, there are two scenarios: planned reuse (e.g. I think this video is awesome and can/should be reused outside), and unplanned reuse (someone really liked that problem I put in unit 4 and the whole thing is CC licensed, so go for it).

The line of questioning we've been going down has mostly been about keeping the Bundle as the versioned increment of authored change and building a layer of more granular reuse inside of it. This makes it easier to work with from an authoring point of view, as we're not having to track a bunch of link dependencies to do simple updates, but it introduces complexities around reuse and sharing because it partly overlaps existing Bundle reuse mechanisms and small in-Bundle file conventions don't necessarily have good isolation guarantees (you can cheat).

But @bradenmacdonald also looked at it from the other side as well -- making Bundles easier to work with in bulk, and making it so that Bundles in the same collection always pointed to the most recent versions of each other. I was thinking about this, when I had a "hold my beer" moment. Because what I've always said about Links and reuse happening at the Bundle/BundleVersion level is not actually true. Reuse and links as envisioned operate at the _BundleSnapshot_ level.

So what if BundleVersions and Snapshots are 1:M instead of 1:1? Each BundleVersion is composed of N Snapshots, so that conceptually BundleVersion = a mapping of names to snapshots. So it kind of gives us directory structure and BlockStore level declaration of dependencies. There are certain ergonomics we could add -- making one Snapshot "special" so that we have a simple default as well as a way to do a simpler v1 cut in a forward compatible way, automatically generated links, etc. If this sounds like I'm back at the Bundle Parts proposal, I sort of am, but I find this way more appealing because it requires fewer new concepts and data modeling to support it.

I need to crash soon, but I just wanted to throw this idea out there before I did. It's entirely possible that I will disavow this entire idea once I've had eight hours of sleep. 😫

ormsbee:
Random side note: If we did the above change and Snapshot becomes a sharable thing instead of an implementation detail, then I'd push for renaming `Snapshot` to `Block`. So a `Bundle` is a set of things that you work on together and it's made of `Blocks` which are the sharable pieces. And then the service name actually makes sense. 😃

ormsbee:
<sigh> Okay, I cannot seem to figure out a way to make that work that isn't more trouble than it's worth in terms of complexity unless we severely restrict how its used. I can sort of see a scenario where all Blocks in a Bundle are peers (e.g. content library) or where there's one top level and a bunch of peers (e.g. Sequences), but it really breaks down if you allow a more varied mixes like Videos.

Maybe Blockstore just stores and recognizes entrypoints that must point to valid files and we reject updates that delete an entrypoint file.  So validation would happen at the Blockstore level.

But I think that doing it this way, individual problems in content libraries should be their own Bundles.

bradenmacdonald:
Hmmm, I'll mull this over.

ormsbee:
Okay, my moment of madness has passed. I think I'm once again comfortable with a world where Bundles are the increment of work and Entry Points (whether a first class Blockstore concept or encoded in a separate file) are the basis of sharing. But I think that we should be careful to frame entry points as a way to share one primary and a few secondary things and not as the basis for sharing very large sets of content, or performance could really suffer.

ormsbee:
> Having Blockstore manage bundle.json may achieve that too. For example, we do send the public attribute when uploading files. That can auto update bundle.json. Though agree that we can review case by case.

Yeah, I think I'm leaning towards having Blockstore manage it now (and maybe not even have an explicit `bundle.json` but fold that into `BundleVersion` metadata).

> To be honest, I am not clear about the purpose of the type field. To me it should be the "type" of the indexable content in the Bundle. And it should be auto-derived by the indexer?

I was thinking mostly as a dispatch system for any plugins, so it knows which plugins to use for a particular Bundle.

> I think at some point we will have to figure out how all this maps to Studio i.e. from the user workflow point of view. The various options have different tradeoffs and it is hard to say which would be a better fit without having some more clarity about that and there is a risk we build out too much that ends up not being the right model if we are not looking at it from a product-first end.

Agreed.

> For example, for LX, the core content XBlocks in units, for example videos, are supposed to show up separately in search results for authors so that they can remix it with their own intro (html XBlock). Does that mean the video XBlock should be in one Bundle, and the unit is supposed to be in another? How would the "commit" work in that case across two Bundles?

I think that in the case where the Bundle is the Sequence and a single user is working on it (i.e. not a separate Video team with their own workflow), we can have both in the same Bundle with different entry points. So we don't worry about cross-bundle commits.

> Another implication is that since keys contain the Bundle id, to move an XBlock out of a Bundle later to make it independently re-usable would mean changing its key and invalidating existing usage_keys. So does that mean that an XBlock can never be made re-usable if the author did not specify that initially (so that it is put in its own Bundle)?

That's a great point. Maybe the usage key should be based on the borrowing Bundle? So if a video is extracted out of a sequence, we're still asking for the usage key relative to that sequence. That opens us up for some name collision issues, but it's conceptually similar to how we'd be thinking of courses and a course context...

bradenmacdonald:
> I have been thinking that we should have an XML level spec for such attributes (title, duration etc) that may need to be displayed/used for filtering but the authoring runtime us simple enough that it may be simpler to just load the XBlocks and call out methods?

The latter is more flexible and is how it works now, I believe (when searching course content with [edx-search](https://github.com/edx/edx-search)): it instantiates the blocks, [calls the `index_dictionary` method](https://github.com/edx/edx-platform/blob/f323ab00ed172cd9af84f920efc42ae27531e7c0/cms/djangoapps/contentstore/courseware_index.py#L168-L187), and then stores that data in the index.

> For example, for LX, the core content XBlocks in units, for example videos, are supposed to show up separately in search results for authors so that they can remix it with their own intro (html XBlock). Does that mean the video XBlock should be in one Bundle, and the unit is supposed to be in another? How would the "commit" work in that case across two Bundles?

I would think that in that case, the remixed *Pathway* is the main thing in the bundle, possibly with the custom intro HTML XBlock also in the same bundle, and everything else is Linked into that bundle. The commit saves the playlist, the html XBlock, and the Links which are all within a single bundle.

> Another implication is that since keys contain the Bundle id, to move an XBlock out of a Bundle later to make it independently re-usable would mean changing its key and invalidating existing usage_keys. 

If that's the case, I think this is a good argument for larger bundles containing multiple things, not small bundles that each contain one thing. (i.e. a Content Library is a Bundle, not a Collection of tiny bundles). But see below, I don't think that's necessarily always the case.

> does that mean that an XBlock can never be made re-usable if the author did not specify that initially (so that it is put in its own Bundle)?

I think it could be nice if we impose the requirement that re-use happens at the granularity defined by the author, so if the author put a bunch of `unit` olx files in a bundle, then they can be re-used like that, but the smaller pieces within them cannot be directly re-used. If the author puts only sequences in a bundle, then only the sequences can be re-used. But in either case, I think it'd be good if the author can later split them up without changing the keys, and that would work with with the current definition key prototype I have, as long as we go for the "multiple entry points per bundle" model instead of the "single entry point per bundle" model.

> That's a great point. Maybe the usage key should be based on the borrowing Bundle? So if a video is extracted out of a sequence, we're still asking for the usage key relative to that sequence. That opens us up for some name collision issues, but it's conceptually similar to how we'd be thinking of courses and a course context...

It's an open question in my mind whether the **definition key** changes when content is borrowed - would love to hear your thoughts on that.

But regarding **usage keys** I was definitely thinking that they'd be relative to where they're used. I was thinking of something like this:

```
bundle 1: "Programming Problems"
/dragon_tail_problem.olx
/birthday_problem.olx

bundle 2: "Braden's Pathway"
/pathway.olx   (a pathway containing an html block intro to the Birthday Problem and an xinclude of the linked problem)
/_links/bundle1/birthday_problem.olx

bundle 3: "Intro to Python Course"
/_links/bundle1/birthday_problem.olx
/outline.xml:
    <course-outline>
        ...
        <section title="Chapter 5">
            <subsection title="Problems">
                <unit include="/_links/bundle1/birthday_problem.olx" usage_id="birthday1"/>
            </subsection>
        </section>
        ...
    </course-outline>
```

With this example, there are a few options:
1. In bundle 2, should the linked "birthday_problem.olx" have the same definition ID as it has when accessed directly in bundle 1? (Easiest answer is "no", I think, but then tracking reuse is harder, and we have to check yet again that no url_names conflict; you couldn't link to a block if its url_name was the same as another block already in the bundle. I'm leaning toward using the same definition ID in both places.)

2. In bundle 3, I'm showing an example of how I'd like to see a new `usage_id` specified whenever a block is used in a course (it can be the same as the `url_name` by default and can be autogenerated by Studio etc.). This means that you can use the same definition in a course in multiple places, but they must have different usage IDs. It also means that whatever form of `UsageKey` we end up using for courses does not have to encode the definition key. If we left the definitionkey out of the course usage keys, then it would be possible to move the original content to another bundle without breaking the usage key, since the usagekey would be based only on the `usage_id` specified in the course, and the definition would be tracked down via xincludes/links.

The key point I'm trying to make here is distinguishing between `DefinitionKeys`, `GlobalUsageKeys` (which in my prototype always encode the definition key), and `CourseUsageKeys` (haven't coded these yet, these are when you use a blockstore block in a course, and they don't necessarily have to encode the definition key).

Let's discuss tomorrow, it might be easier to explain on a call.

ormsbee:
@bradenmacdonald: I need to write a bunch of tests and rebase onto the mysql changes, but I'd appreciate it if you could take a first pass at the review. The two things we discussed yesterday that are not here yet are:

1. Partial commits -- there's support at the data layer and a clear place for it in the REST API, but I haven't hooked it up yet since it didn't seem critical.
2. Presenting a Bundle's Drafts as a dict instead of a list. Mostly because I spun my wheels with serializers trying to figure out the right way to make that work and I'm punting on it.

Otherwise, the data model changes and basic flow of Draft creation, updates, and commits should be working.

blockstore/apps/bundles/store.py
bradenmacdonald: I think it's also worth mentioning that some django-storages backends have an option to allow direct overwrites via `save`, e.g. https://github.com/jschneier/django-storages/search?q=file_overwrite&unscoped_q=file_overwrite

bradenmacdonald:
@ormsbee Do you have any instructions / tips on how to test this manually? I remember you showing us a demo over Google Hangout but from a quick look now I'm not seeing how to upload files into a draft etc. via the web UI.

ormsbee:
Ah, good point. I’ll put that in.

blockstore/apps/api/v1/serializers/drafts.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class DraftFileUpdateSerializer(serializers.BaseSerializer):
+    """Write-only serializer for Draft file updates."""
+
+    def to_internal_value(self, data):
+        """
+        Convert file names -> b64 strings into file names -> ContentFiles.
+
+        `data` is a  dict of file names to base64 encoded data representations.
```
bradenmacdonald: How does one delete a file from a draft? (so it exists in the bundle, but I want it to not exist once the draft is committed)

How does one reset a files changes in the draft? (so the file now has `"modified": false` but doesn't have to be re-uploaded via `PATCH`. Fine if this is added later, but I think deleting files is needed now.

blockstore/apps/api/v1/views/drafts.py
```diff
@@ -0,0 +1,179 @@
...
+                request=request,
+            )
+        }
+        return Response(result, status=status.HTTP_201_CREATED)
+
+    def destroy(self, request, uuid):
+        """
+        This removes any files that were staged along with the database entry.
+        """
+        draft_repo = DraftRepo(SnapshotRepo())
```
bradenmacdonald: Nit: `draft_repo = DraftRepo(SnapshotRepo())` comes up a lot; is there a reason not to make it a long-lived/global singleton or module-level var? Or make a property access to get it as `self.draft_repo` ?

blockstore/apps/api/v1/views/snapshots.py
```diff
@@ -1,18 +1,66 @@
...
-from ...constants import FILE_PATH_REGEX
+from ...constants import FILE_PATH_REGEX, SNAPSHOT_HASH_DIGEST, UUID4_REGEX
 from ...utils import ZippableMultiValueDict
-from ..serializers.snapshots import FileInfoSerializer
+from ..serializers.snapshots import FileInfoSerializer, SnapshotSerializer
+
+
+class SnapshotViewSet(viewsets.ViewSet):
+    """
+    The Snapshots resource can only be accessed in a read-only detail view.
```
bradenmacdonald: Is there an example use case where, say, Studio might want to query a snapshot's details via this API? Or is this mostly for helping with blockstore implementation?

blockstore/apps/bundles/admin.py
```diff
@@ -51,21 +54,32 @@ def snapshot_data(self, obj):
...
 
+class DraftInline(admin.StackedInline):
+    """Draft inline view. """
+    model = Draft
+    classes = ['collapse']
+
+    def get_max_num(self, request, obj=None, **kwargs):
+        if obj is None:
+            return 3
+        return max(obj.drafts.count(), 3)
```
bradenmacdonald: This seems odd to me. Are there scenarios where one might want to create multiple drafts at the same time?

blockstore/apps/bundles/store.py
```diff
@@ -208,68 +203,128 @@ class BundleDataStore():
...
+    def get(self, bundle_uuid: UUID, snapshot_digest: bytes) -> Snapshot:
+        """
+        Return a snapshot.
 
-        return BundleSnapshot(
+        TODO: There's no meaningful error handling yet.
+        """
+        storage_path = self._summary_path(bundle_uuid, snapshot_digest)
+        try:
+            with self.storage.open(storage_path) as snapshot_file:
```
bradenmacdonald: Storing this summary file on S3 rather than MySQL still seems very odd to me, but let's see how it goes :)

blockstore/apps/bundles/store.py
```diff
@@ -208,68 +203,128 @@ class BundleDataStore():
...
+    def url(self, snapshot, path):
+        """Return a user-accessible URL to download a path from this Snapshot."""
+        file_info = snapshot.files[path]  # pylint: disable=unsubscriptable-object
+        storage_path = self._file_data_path(snapshot.bundle_uuid, file_info.hash_digest.hex())
+        return self.storage.url(storage_path)
+
+    def open(self, snapshot, path):
+        file_info = snapshot.files[path]  # pylint: disable=unsubscriptable-object
+        storage_path = self._file_data_path(snapshot.bundle_uuid, file_info.hash_digest.hex())
+        return self.storage.open(storage_path, 'rb')
```
bradenmacdonald: What are the circumstances under which we'd read file contents within a `SnapshotRepo`? I think we should avoid reading file contents from the repo like the plague for performance reasons (only ever returning URLs and letting clients read file content directly from S3/CloudFront). Obviously we need the hash on the initial upload, but this method is for reading the content of files that have already been uploaded into the repo.

From what I can tell, this is only used (via `file_mapping`) in `DraftRepo.commit()`, but it "feels" to me like it should be possible to commit a draft without Blockstore needing to read any file contents. We already know the hash and size of every file in the draft.

blockstore/apps/bundles/store.py
```diff
@@ -208,68 +203,128 @@ class BundleDataStore():
...
+
+    @classmethod
+    def _file_data_path(cls, bundle_uuid, file_hash):
+        return f'{bundle_uuid}/snapshot_data/{file_hash}'
+
     def _save_file(self, bundle_uuid, path, data, public=False):
         """
         Save file at path and return a FileInfo object for it.
         """
+        # import pudb; pu.db
```
bradenmacdonald: Nit: stray comment?

blockstore/apps/bundles/store.py
```diff
@@ -279,28 +334,310 @@ def _create_snapshot(self, bundle_uuid, files):
...
+    def _data_file_path(cls, draft_uuid, file_path):
+        """Path to a file that we've edited in our Draft."""
+        return f'{draft_uuid}/data/{file_path}'
+
+    @classmethod
+    def _summary_path(cls, draft_uuid):
+        return f'{draft_uuid}/summary.json'
+
+    def _overwrite(self, path, file_obj):
+        # There's gotta be a better way, but for now...
```
bradenmacdonald: I mentioned this in reply to a comment on an older version of this, but for since that mention is now hidden, I'll re-post it here where it's still relevant:

> some django-storages backends have an option to allow direct overwrites via `save`, e.g. https://github.com/jschneier/django-storages/search?q=file_overwrite&unscoped_q=file_overwrite

I believe it may also be relatively trivial to subclass django's built-in filestorage implementation to support overwriting for filesystem storage. [django-storages used to include exactly that](https://django-storages.readthedocs.io/en/1.2/backends/overwrite.html) but for some reason no longer does.

blockstore/apps/api/constants.py
```diff
@@ -2,6 +2,11 @@
 
 FILE_PATH_REGEX = '[^:]+'
 
+# Collections, Bundles, and Drafts all use UUIDs
 UUID4_REGEX = '[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}'
```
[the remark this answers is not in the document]
ormsbee: Honestly, is there anything keeping us from using Django 2.1 now? Django 2.2 is the next LTS, and it's due to drop in April.

blockstore/apps/api/v1/serializers/drafts.py
```diff
@@ -0,0 +1,109 @@
...
+
+    This payload could become very large, so this serializer should not be used
+    for list views.
+    """
+    class Meta:
+        model = Draft
+        queryset = Draft.objects.all()
+        fields = DraftSerializer.Meta.fields + ('staged_draft',)
+
+    staged_draft = StagedDraftField()
```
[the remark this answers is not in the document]
ormsbee: Yeah, this was an artifact of data model refactoring over time. I'll consolidate it.

blockstore/apps/api/v1/views/drafts.py
```diff
@@ -0,0 +1,179 @@
...
+
+        PATCH is a bit unusual for this resource in that the only thing you can
+        patch are file data. You cannot change the draft name or the Bundle it
+        belongs to, and the only way to update the `base_snapshot` is to commit
+        the Draft.
+
+        There is intentionally no PUT support for file data, for a few reasons:
+
+        1. We can't guarantee the semantics of a PUT against a concurrent PATCH,
+        particularly for large numbers of files. Our files are in an object
```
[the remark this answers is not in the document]
ormsbee: Hah. Yeah. I still think PATCHES will get us into less trouble, but... let me think about that one. Will probably punt until after this PR though.

blockstore/apps/api/v1/views/drafts.py
```diff
@@ -0,0 +1,179 @@
...
+        multiple requests to make the necessary changes.
+
+        3. It's just simpler to have only one way to update the files.
+        """
+        serializer = DraftFileUpdateSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        files_to_write = serializer.validated_data
+
+        draft_repo = DraftRepo(SnapshotRepo())
+        draft_repo.update_files(uuid, files_to_write)
```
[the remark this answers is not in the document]
ormsbee: Yeah, we'll have to do that.

blockstore/apps/api/v1/serializers/drafts.py
ormsbee: Delete is PATCHing with that filename set to null.

No facility for resetting at the moment.

blockstore/apps/api/v1/views/snapshots.py
ormsbee: But yeah, it's useful in the short term just for implementation. I'm not averse to stripping it out afterwards if we don't end up using it.

blockstore/apps/bundles/admin.py
ormsbee: I think that imports will want their own private draft space.

blockstore/apps/api/constants.py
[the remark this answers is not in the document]
bradenmacdonald: > is there anything keeping us from using Django 2.1 now?

Nothing I know of.

blockstore/apps/bundles/admin.py
bradenmacdonald: Right, I just meant that I don't foresee someone manually creating Studio + Import draft spaces from the django admin UI at the same time.

blockstore/apps/api/v1/views/snapshots.py
ormsbee: You know what, you're right. I'm just going to remove this.

blockstore/apps/api/v1/serializers/drafts.py
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        basic_info['files'] = {
+            path: {
+                "url": draft_repo.url(staged_draft, path),
+                "size": file_info.size,
+                "hash_digest": file_info.hash_digest.hex(),
+                "modified": path in staged_draft.files_to_overwrite
+            }
+            for path, file_info in staged_draft.files.items()
```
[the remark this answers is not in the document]
ormsbee: Killing the Snapshot files serializer (along with its ViewSet).

blockstore/apps/bundles/store.py
ormsbee: Oops...

blockstore/apps/bundles/store.py
ormsbee: I'll try the subclassing, but my concern would be that overwrites are in this case necessary for the application to function properly at all. So if we code with the assumption that overwrite will be enabled on any particular backend, then people can shoot themselves in the foot by not properly configuring their storage bucket system. AWS defaults to overwrites = True, but Azure doesn't, etc.

blockstore/apps/bundles/store.py
ormsbee: So I was mucking with this a bit last night...

I agree that opening files should be avoided, and I like the idea of not even allowing an interface for that. The original reason I put this in is because I wanted some kind of barrier between Drafts and Snapshots, because I feel a lot more uncertainty about Drafts in the long term, and I'd like to have the flexibility to keep it more separate (different backing store, different whatever) so that it's easier to toss aside and replace. I didn't want to cram in a formal separation by app in this PR because it's already way too big and been open way too long, but I figured that later on it would end up in that direction.

On the other hand, the only way to make it so that we can guarantee that `commit()` is a fast, synchronous operation is if drafts share the same backing store and we can quickly do moves of files instead of rewriting them. But unfortunately, moves don't exist in the Django File Storage API. We'd have to dig down to lower level abstractions and have our own custom backends for AWS and other backends if we wanted to maintain that.

So is it okay if we leave this in here for now, and try to figure out the bigger story about the layer of abstraction for Drafts becoming Snapshots afterwards?

blockstore/apps/bundles/store.py
ormsbee: Going to punt this since it's likely we'll go with custom backends entirely to enable synchronous commits.

blockstore/apps/bundles/store.py
bradenmacdonald: I see. Yep, as discussed, I'm keen to get this merged so we can figure that out in a future PR.

blockstore/apps/api/v1/tests/test_contract.py
```diff
@@ -0,0 +1,384 @@
...
+class CollectionsTestCase(TestCase):
+    """Test basic Collections CRUD operations."""
+
+    def setUp(self):
+        super().setUp()
+        self.client = APIClient()
+
+    def test_list_of_one(self):
+        """
+        Make sure list[0] and detail have the same representation.
```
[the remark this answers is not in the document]
ormsbee: Yeah, nice catch. I'll make that change.

blockstore/apps/api/v1/tests/test_contract.py
```diff
@@ -0,0 +1,384 @@
...
+        assert draft_data['name'] == 'studio_draft'
+        assert 'staged_draft' not in draft_data
+
+        # Getting the result from a retrieve (GET on individual Draft) should
+        # yield the same data, except that it should *also* contain the
+        # metadata about the files inside (staged_draft)
+        detail_response = self.client.get(draft_data['url'])
+        detail_data = response_data(detail_response)
+        for key, value in detail_data.items():
+            assert detail_data[key] == value
```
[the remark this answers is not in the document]
ormsbee: Yup, ty!

blockstore/apps/bundles/store.py
```diff
@@ -208,68 +210,127 @@ class BundleDataStore():
...
+        """
+        Save the files, create a Snapshot object and save its JSON serialization to storage.
+        """
+        files = {}
+        for path, file in paths_to_files.items():
+            files[str(path)] = self._save_file(bundle_uuid, path, file)
+
+        return self._create(bundle_uuid, files)
+
+    def create_by_adding_paths(self, snapshot, paths_to_files):
```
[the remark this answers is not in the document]
ormsbee: Good point. Removing...

blockstore/apps/bundles/store.py
```diff
@@ -279,28 +340,348 @@ def _create_snapshot(self, bundle_uuid, files):
...
+
+        if path in draft.files_to_overwrite:
+            data_file_path = self._data_file_path(draft.uuid, path)
+            return self.storage.url(data_file_path)
+
+        return self.snapshot_repo.url(draft.base_snapshot, path)
+
+
+class BundleDataJSONEncoder(json.JSONEncoder):
+    """Default JSON serialization."""
```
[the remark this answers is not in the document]
ormsbee: Good point. I agree it should happen in a followup.

ormsbee:
My personal goal for this project is to never have a PR this long and drawn out again. 😝

PR #14, opened 2018-11-06
Update Tagstore to work with MySQL, Blockstore to work with utf8mb4

bradenmacdonald:
When I rebased https://github.com/open-craft/blockstore/pull/7 in preparation for merge, a couple tests broke (specifically, emoji tags weren't working and case-sensitive tags + case-insensitive search was not working). Here's what eventually I arrived at to fix them.

This PR:
* Uses a migration to force Blockstore's MySQL database to use the `utf8mb4` encoding and `utf8mb4_unicode_ci` collation. These are generally recommended by everyone for new projects, and we already had some design decisions in place to be compatible with those. (However, without the fixes in this PR, it wasn't actually possible to use blockstore with `utf8mb4`)
* Changes tagstore from being case-sensitive for tags to "preserve case but tags must be case-insensitively unique". I think that's better anyways, and it doesn't matter because on MySQL (and only MySQL) django doesn't allow you to do case-insensitive searches on a column that has case-sensitive collation.

Testing instructions:

1. Drop all tables in your blockstore database, or re-create it, or run all django app migrations to `zero`.
1. Check out this branch.
1. Run `make blockstore-shell` then `./manage.py migrate`
1. Run `make test` to confirm the tests pass.

### DIFF
--- a/blockstore/apps/mysql_unicode/__init__.py
+++ b/blockstore/apps/mysql_unicode/__init__.py
@@ -1,0 +1,5 @@
+"""
+An app to force use of the non-broken Unicode character encoding with MySQL.
+
+See migrations/0001_initial.py for details.
+"""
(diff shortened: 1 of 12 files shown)


bradenmacdonald:
@ormsbee are you cool with this? Or is there a better way?

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5477](https://openedx.atlassian.net/browse/OSPR-5477) so that we can track it in Jira.

PR #15, opened 2018-11-07
Tagstore API improvements

bradenmacdonald:
A few last API improvements to make this tagstore API a bit nicer, before we get too locked in.

* Per @symbolist's suggestion, `Tag.tag` is now `Tag.name` which I think is much nicer
* The API was missing a way to get an existing `Tag` without constructing it yourself, so I added a `get_tag()` method
* Made the `Taxonomy` class a bit more pythonic by giving it some methods, so now you can do `taxonomy.add_tag('foo')` instead of `tagstore.add_tag_to_taxonomy('foo', taxonomy)`

Testing instructions:
* The unit tests should cover everything quite well, so mostly review the code. You can also run the updated example code from `README.rst`.

### DIFF
--- a/tagstore/README.rst
+++ b/tagstore/README.rst
@@ -13,7 +13,7 @@ Features
 * Python 3 API with type hints
 * Allows any "entity" to be tagged, where an entity could be a user, a block, a collection, etc.
 * Allows rich searching for entities by tags. e.g. "Find all large animals" will return an entity that was tagged with "large" and "dog", since it knows that the "dog" tag is a type of "animal" tag.
-* Designed to support multiple tag storage backends, although the current version of Tagstore only includes a Django ORM backend, which stores tags in MySQL/PostgeSQL/SQLite. (A reasonably complete Neo4j backend implementation also existed in an early version and can found at https://github.com/open-craft/blockstore/pull/7/commits/05503ade0d29296119578837fe059a04e57209e0)
+* Designed to support multiple tag storage backends, although the current version of Tagstore only includes a Django ORM backend, which stores tags in MySQL/PostgeSQL/SQLite. (A reasonably complete Neo4j backend implementation also existed in an early version and can found at https://github.com/open-craft/blockstore/commit/714b22f8456fb3b509aca54f59dc96de060d36fe)
 
 Non-features
 ------------
@@ -40,6 +40,18 @@ Here is an example of using the Tagstore API::
     pine = biology.add_tag('pine', parent_tag=conifer)
     aster = biology.add_tag('aster', parent_tag=plant)
 
+    # Print tag hierarchy tree:
+    depths = {}
+    for (tag, parent) in biology.list_tags_hierarchically():
+        depths[tag] = depths[parent] + 1 if parent else 0
+        print(("  " * depths[tag]) + tag.name)
+    # The resulting hierarchy that gets printed out is:
+    #   plant
+    #     aster
+    #     conifer
+    #       cypress
+    #       pine
+
     # Create a "sizes" taxonomy:
     sizes = tagstore.create_taxonomy("sizes", owner_id=None)
     small = sizes.add_tag('small')
@@ -55,29 +67,29 @@ Here is an example of using the Tagstore API::
     tagstore.add_tag_to(cypress, redwood)
 
     # Find all asters
-    set([e for e in tagstore.get_entities_tagged_with(aster)])
+    set(tagstore.get_entities_tagged_with(aster))
     # result: {dandelion}
 
     # plants
-    set([e for e in tagstore.get_entities_tagged_with(plant)])
+    set(tagstore.get_entities_tagged_with(plant))
     # result: {dandelion, redwood}
 
     # small plants
-    set([e for e in tagstore.get_entities_tagged_with_all({plant, small})])
+    set(tagstore.get_entities_tagged_with_all({plant, small}))
     # result: {dandelion}
 
     # plants, with no tag inheritance
-    set([e for e in tagstore.get_entities_tagged_with(plant, include_child_tags=False)])
+    set(tagstore.get_entities_tagged_with(plant, include_child_tags=False))
     # result: set()
 
     # conifers
-    set([e for e in tagstore.get_entities_tagged_with(conifer)])
+    set(tagstore.get_entities_tagged_with(conifer))
     # result: {redwood}
 
-    # small things starting with "d"
-    set([e for e in tagstore.get_entities_tagged_with(
-        small, entity_types=['thing'], external_id_prefix='d'
-    )])
+    # plants starting with "d"
+    set(tagstore.get_entities_tagged_with(
+        plant, entity_types=['thing'], external_id_prefix='d'
+    ))
     # result: {dandelion}
 
 
--- a/tagstore/README.rst
+++ b/tagstore/README.rst
@@ -34,17 +34,17 @@ Here is an example of using the Tagstore API::
 
     # Create a biology taxonomy:
     biology = tagstore.create_taxonomy("Biology", owner_id=None)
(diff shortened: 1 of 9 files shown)


tagstore/README.rst
```diff
@@ -34,17 +34,17 @@ Here is an example of using the Tagstore API::
...
+    aster = biology.add_tag('aster', parent_tag=plant)
 
     # Create a "sizes" taxonomy:
     sizes = tagstore.create_taxonomy("sizes", owner_id=None)
-    small = tagstore.add_tag_to_taxonomy('small', sizes)
-    med = tagstore.add_tag_to_taxonomy('med', sizes)
-    large = tagstore.add_tag_to_taxonomy('large', sizes)
+    small = sizes.add_tag('small')
+    med = sizes.add_tag('med')
+    large = sizes.add_tag('large')
```
pomegranited: Since we're nitting, the [`set` examples below](https://github.com/open-craft/blockstore/pull/15/files#diff-68fd5bf025c3883a22f4b583aa68fd56R58) can be shortened to:
```diff
diff --git a/tagstore/README.rst b/tagstore/README.rst
index dc9f22a..267af95 100644
--- a/tagstore/README.rst
+++ b/tagstore/README.rst
@@ -55,29 +55,29 @@ Here is an example of using the Tagstore API::
     tagstore.add_tag_to(cypress, redwood)
 
     # Find all asters
-    set([e for e in tagstore.get_entities_tagged_with(aster)])
+    set(tagstore.get_entities_tagged_with(aster))
     # result: {dandelion}
 
     # plants
-    set([e for e in tagstore.get_entities_tagged_with(plant)])
+    set(tagstore.get_entities_tagged_with(plant))
     # result: {dandelion, redwood}
 
     # small plants
-    set([e for e in tagstore.get_entities_tagged_with_all({plant, small})])
+    set(tagstore.get_entities_tagged_with_all({plant, small}))
     # result: {dandelion}
 
     # plants, with no tag inheritance
-    set([e for e in tagstore.get_entities_tagged_with(plant, include_child_tags=False)])
+    set(tagstore.get_entities_tagged_with(plant, include_child_tags=False))
     # result: set()
 
     # conifers
-    set([e for e in tagstore.get_entities_tagged_with(conifer)])
+    set(tagstore.get_entities_tagged_with(conifer))
     # result: {redwood}
 
     # small things starting with "d"
-    set([e for e in tagstore.get_entities_tagged_with(
-        small, entity_types=['thing'], external_id_prefix='d'
-    )])
+    set(tagstore.get_entities_tagged_with(
+        plant, entity_types=['thing'], external_id_prefix='d'
+    ))
     # result: {dandelion}
```

(and changing the final `get_entities_tagged_with` to use the `plant` tag instead of the `small` tag better illustrates the `external_id_prefix`, since `{dandelion, redwood}` are tagged with `plant`).

tagstore/backends/django.py
```diff
@@ -24,52 +24,64 @@ def create_taxonomy(self, name: str, owner_id: Optional[UserId]) -> TaxonomyMeta
...
 
-    def _add_tag_to_taxonomy(self, taxonomy_uid: int, tag: str, parent_tag: Optional[str] = None) -> str:
+    def _add_tag_to_taxonomy(self, taxonomy_uid: TaxonomyId, name: str, parent_tag: Optional[str] = None) -> str:
         if parent_tag:
             # Check the parent tag:
             try:
-                pt = TagModel.objects.get(taxonomy_id=taxonomy_uid, tag=parent_tag)
+                pt = TagModel.objects.get(taxonomy_id=taxonomy_uid, name=parent_tag)
             except TagModel.DoesNotExist:
                 raise ValueError("Invalid parent tag.")
```
pomegranited: If you provide tests for these two lines, [41-42](https://github.com/open-craft/blockstore/pull/15/files#diff-b4fdc37c88d35346e1890617408283a8R41), and [141-142](https://github.com/open-craft/blockstore/pull/15/files#diff-b4fdc37c88d35346e1890617408283a8R141) below, then your coverage for the `tagstore` dir will reach 100%.

tagstore/models/taxonomy.py
```diff
@@ -1,16 +1,68 @@
...
     name: str
     owner_id: Optional[UserId]
+
+    tagstore: Any  # Type is Tagstore but we can't define that without circular import
+
+    # Convenience methods:
+
+    def add_tag(self, name: str, parent_tag: Optional[Tag] = None) -> Tag:
+        """
+        Add the specified tag to this given taxonomy, and retuns it.
```
pomegranited: nit: `retuns` should be `return` (fix type and verb plurality).

tagstore/models/taxonomy.py
```diff
@@ -1,16 +1,68 @@
...
+
+        Will be a no-op if the tag already exists in the taxonomy (case-insensitive),
+        however the returned (existing) Tag may differ in case.
```
pomegranited: I wasn't able to find a case where the "returned (existing) Tag may differ in case" occurrs..  suggest something like this instead?

> If a Tag already exists in the taxonomy with the given name (case-insensitive) with the given parent, then that Tag is returned and no changes are made.

tagstore/README.rst
pomegranited: Worth also mentioning the list method too:
```python
# get tag hierarchy
list((tag.name, parent.name) for (tag, parent) in biology.list_tags_hierarchically())
# result: [('plant', None), ('aster', 'plant'), ('conifer', 'plant'), ('cypress', 'conifer'), ('pine', 'conifer')]
```

tagstore/tagstore.py
```diff
@@ -74,14 +64,22 @@ def _add_tag_to_taxonomy(self, taxonomy_uid: int, tag: str, parent_tag: Optional
         """
         raise NotImplementedError()
 
-    def list_tags_in_taxonomy(self, uid: int) -> Iterator[Tag]:
+    def get_tag_in_taxonomy(self, name: str, taxonomy_uid: TaxonomyId) -> Optional[Tag]:
```
pomegranited: Docstring for [`_add_tag_to_taxonomy` above](https://github.com/open-craft/blockstore/pull/15/files#diff-a649213e62430d6f8f65bc09c1007e05R62) should read:

> It should return the 'name' value of the newly created tag, ...

tagstore/models/taxonomy.py
bradenmacdonald: > I wasn't able to find a case where the "returned (existing) Tag may differ in case" 

There is a test case showing that? https://github.com/open-craft/blockstore/blob/05a88e283d81f67bb7094df6d7a453ef40ab3890/tagstore/backends/tests.py#L73-L75

I do like your wording though, so I'll use it.

tagstore/README.rst
bradenmacdonald: I did one better (I think?):

```python
    # Print tag hierarchy tree:
    depths = {}
    for (tag, parent) in biology.list_tags_hierarchically():
        depths[tag] = depths[parent] + 1 if parent else 0
        print(("  " * depths[tag]) + tag.name)
    # The resulting hierarchy that gets printed out is:
    #   plant
    #     aster
    #     conifer
    #       cypress
    #       pine
```

bradenmacdonald:
Thanks @pomegranited, great points. Fixes are all made in aba94c9dedb013eb35b5785567fa9877960345f3

tagstore/models/taxonomy.py
pomegranited: Looks like your change to the wording hasn't make it in yet?

bradenmacdonald: It did make it in in `tagstore.py` but I missed `taxonomy.py`. Fixed now :)

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5476](https://openedx.atlassian.net/browse/OSPR-5476) so that we can track it in Jira.

PR #17, opened 2018-11-14
[LX-113] Django admin tool for managing taxonomies

SSPJ:
## Description

This PR concerns the creation of a graphical UI for adding tags. Via the tagstore API, it can:
- create/edit/delete taxonomies (where the deletion of a taxonomy deletes all child tags)
- add nested tags to a taxonomy and render a hierarchical display of those tags

Via the Django OEM, it can:
- create/edit/delete & add tags to Entities

## Manual Test Plan

Pull down `seamus/tag-admin` and `make runserver` as normal. Visit http://0.0.0.0:18250/admin/tagstore_django/ in the browser. Add a Taxonomy, add tags to it, delete the taxonomy. Also create an Entity and test tagging it. Delete the Entity.

## Testing Checklist
- [x]: Include unit, integration, acceptance tests as appropriate
- [x]: Include accessibility (a11y) tests: I tested with my keyboard
- TODO: Include tests that capture the external-query scaling properties
- [ ] Manually test right-to-left languages and i18n of the changes.

## Non-testing Checklist
- __n/a?__ Tag DevOps on your PR (user `edx/devops`), either for review
  (more specific to devops = better) or just a general FYI.
- [x] Consider any documentation your change might need, and which
  users will be affected by this change.
- [x] Double-check that your commit messages will make a meaningful release note.

## Post-review
- TODO: Squash commits into discrete sets of changes (see the note about release notes above)

## Reviewers
- [ ] Code review #1 @bradenmacdonald 
- [ ] Code review #2 (TODO: tag a specific user)
- [ ] Docs review (required for UI strings/error messages)
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

### Areas to Consider
- [ ] i18n 
    - Are all user-facing strings tagged?
        - They all have trans on them.
- [ ] Right-to-left
    - Will the feature work for right-to-left languages?
- [ ] Analytics
    - Are any new events being emitted?
        - The tagstore API should handle analytics for stuff created through it.
    - Have any events been changed?
    - Are there any new user actions that should be emitted as events?
- [ ] Performance
    - What dimensions does this change scale over? Users? Courses? Course size?
    - How does the feature scale over those dimensions? Sub-linear? Linear? Quadratic? Exponential?
    - How does the scaling affect the number of external calls (database queries,
      api requests, etc) that this code makes?
        - It should be able to handle several thousand tags without issue. Beyond that, a different UI design might be better.
- [ ] Database migrations
    - Are they backwards compatible?
        - ?
    - When they run on production, how long will they take? Will they lock the table?
        - They should be instantaneous and no.


bradenmacdonald:
>  Include accessibility (a11y) tests -- How do I do these? edx accessibility docs doesn't say anything about programmatic tests

We don't usually worry about a11y in the django admin too much, since it's not used by end users, and we definitely don't need to test for it.

Migrations being backwards compatible just means that the migrations will apply cleanly on top of any existing migrations. Since blockstore isn't yet used in production anywhere, we don't really care about that (yet) either.

In fact, you can ignore the whole "Testing checklist" here as it doesn't really apply to this project (or at least at this stage). It's a cookie-cutter template from edX.

blockstore/templates/admin/tagstore_django/list_item.html
```diff
@@ -0,0 +1,19 @@
+{# Template to display a hierarchical list of tags on the admin page #}
+{% load i18n %}
+
+{% for tag in taglist %}
+		<li>
```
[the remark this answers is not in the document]
SSPJ: Nice catch, thanks! Atom didn't guess the indentation type correctly. I've set it to always soft tab, so this shouldn't crop up again. (Well, until I go to edit a Makefile :) )

tagstore/backends/django.py
```diff
@@ -79,6 +79,33 @@ def list_tags_in_taxonomy_hierarchically(self, taxonomy_uid: TaxonomyId) -> Iter
         for tag in TagModel.objects.filter(taxonomy_id=taxonomy_uid).order_by('path'):
             yield (Tag(taxonomy_uid=taxonomy_uid, name=tag.name), tag.parent_tag_tuple)
 
+    def list_tags_in_taxonomy_hierarchically_as_dict(self, taxonomy_uid: TaxonomyId) -> dict:
+        """
+        Get all tags in the given taxonomy as a list of nested dictionaries.
+
+        Returns a list in the form of [tag, tag, ...] where tag is a dictionary
+        containing the name and id of a top level tag (it has no parent)
+        and also a list holding all `children` of that tag recursively (can be empty).
```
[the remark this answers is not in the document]
SSPJ: Did so, and changed the function name to be clearer. Let me know if it's better.

tagstore/backends/tagstore_django/migrations/0002_auto_20181112_1955.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('tagstore_django', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='taxonomy',
+            name='owner',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to='tagstore_django.Entity'),
```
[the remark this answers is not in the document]
SSPJ: Have done so.

tagstore/backends/django.py
```diff
@@ -79,6 +79,33 @@ def list_tags_in_taxonomy_hierarchically(self, taxonomy_uid: TaxonomyId) -> Iter
...
+                stack.append(node)
+            elif parent == stack[-1]['name']:
+                stack[-1]['children'].append(node)
+                stack.append(node)
+            else:
+                while parent != stack[-1]['name']:
+                    stack.pop()
+                stack[-1]['children'].append(node)
+                stack.append(node)
+        return tags
```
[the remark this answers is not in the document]
SSPJ: Thank you for this. That use of None as a key to represent the no-parent top level tags is clever!

The `id` is necessary for constructing change/delete links Django's admin, unfortunately. We're not using it right now, but I anticipate we will want to later.

SSPJ:
Thanks for what you've reviewed so far! I've addressed what you've brought up; feel free to point out any other issues you want looked at.

I've implemented the ability to create/delete/tag Entity models through the UI. There are other ways to make this nicer, for example, refactoring the `_add_tag_to_taxonomy` method to take a `dry_run` flag so that it can be used for validation (as it is, the UI pops up a confusing 'success' toast, even when the tag didn't save), but I will let you decide if more needs done here.

tagstore/backends/tagstore_django/models.py
```diff
@@ -44,7 +47,7 @@ class Taxonomy(models.Model):
     """
     id = models.BigAutoField(primary_key=True)
     name = models.CharField(max_length=MAX_CHAR_FIELD_LENGTH)
-    owner = models.ForeignKey(Entity, null=True, on_delete=models.SET_NULL)
+    owner = models.ForeignKey(Entity, null=True, blank=True, on_delete=models.SET_NULL)
 
     class Meta:
         db_table = 'tagstore_taxonomy'
```
bradenmacdonald: Can you please add `verbose_name_plural = 'Taxonomies'` to this `Meta` class? Otherwise it shows up in the admin as "Taxonomys"

tagstore/backends/tagstore_django/admin.py
```diff
@@ -0,0 +1,80 @@
...
+            tagstore.add_tag_to_taxonomy(name, taxonomy.id, parent)
+        except ValueError:
+            pass
+
+    # def delete_model(self, request, obj):
+    #     """ TODO: Uses the tagstore API to delete tags from the database. """
+    #     super(TagAdmin, self).delete_model(request, obj)
+
+
+class TagInline(admin.TabularInline):
```
bradenmacdonald: Please just remove this for now if we're not using it yet, or at least comment it out.

tagstore/backends/tagstore_django/models.py
bradenmacdonald: Same with "Entities" vs. "Entitys" actually.

tagstore/backends/tagstore_django/admin.py
```diff
@@ -0,0 +1,80 @@
...
+        self.fields['parent'].widget.attrs.update({'value': parent_tag})
+
+    # TODO: create a dry run flag for add_tag_to_taxonomy which
+    # can be used to validate before attempting to save
+
+
+class TagAdmin(admin.ModelAdmin):
+    """ Controls display and saving of Tag model objects. """
+    list_filter = ('taxonomy',)
+    list_display = ('name', 'path', 'taxonomy')
```
bradenmacdonald: I can't currently access the `list` views, so these don't seem to be having any effect?

SSPJ: I wasn't sure the project's stance on avoiding dead code vs leaving useful scaffolding for features we'll likely want later, so thanks for clarifying. I have taken that out.

SSPJ:
@bradenmacdonald I agree with you re: the scalability and I've made all the changes requested.

I ran into a strange issue where the code in `get_tags_in_taxonomy_hierarchically_as_dict` was raising a key error. When I looked at with pdb, it seemed that Tag() and tag.parent_tag_tuple were returning different hashs. Since it was working before, I can only guess that something changed between 3.6.6 and 3.6.7. I think it is related to whether taxonomy_uid is a string or an int, but decided not to chase it further.

bradenmacdonald:
@SSPJ  That hash issue is worrying... I'd rather we identify the underlying cause. Otherwise we might end up with subtle bugs in the future.

I also noticed:

In the admin there is no way to add a new root tag to a taxonomy - only a child tag.

The migration needs to be updated again due to the new verbose name `Meta` options on the models.

tagstore/backends/tagstore_django/templatetags/tagstore_admin.py
```diff
@@ -2,21 +2,13 @@
...
 
 register = template.Library()
 
 
 @register.inclusion_tag('admin/tagstore_django/tag_hierarchy.html')
-def tag_hierarchy():
-    """ Renders a hierarchical view in HTML of all Tag objects. """
+def tag_hierarchy(taxonomy_uid):
+    """ Renders a hierarchical view in HTML of Tag objects. """
     tagstore = DjangoTagstore()
```
bradenmacdonald: @SSPJ I figured out the issue: `taxonomy_uid` is coming in as a string here. If you add `taxonomy_uid = int(taxonomy_uid)` to this method before calling `get_tags_in_taxonomy_hierarchically_as_dict`, you won't get the `KeyError` anymore.

SSPJ:
After casting `taxonomy_uid` to an int, I had no more trouble. Thanks for spotting the issue! I put the cast in `get_tags_in_taxonomy_hierarchically_as_dict` itself to make the method a bit more robust.

> In the admin there is no way to add a new root tag to a taxonomy - only a child tag.

It's accessible via the add button on http://0.0.0.0:18250/admin/tagstore_django/ , but you are right: that is neither intuitive or convenient. I've added a link in the tag hierarchy to address this.

Sorry about the migrations. I've done them now.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5475](https://openedx.atlassian.net/browse/OSPR-5475) so that we can track it in Jira.

PR #18, opened 2018-11-15
Changes to settings and requirements file required for deployment

pkulkark:
## Description
This PR fixes some of the errors seen during automated deployment. The errors were related to missing keys in logging object, changes in syntax from python2 to python3 and additional packages required to apply django migrations.

## Testing Checklist
- [ ] Manually test right-to-left languages and i18n
  of the changes.

## Post-review
- TODO: Squash commits into discrete sets of changes (see the note about release notes above)

## Reviewers
- [ ] @bradenmacdonald

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ requirements-test: ## Install requirements for testing
 	${VENV_BIN}/pip install -qr requirements/test.txt --exists-action w
 
 production-requirements:
-	pip install -r requirements.txt --exists-action w
+	pip install -r requirements/production.txt --exists-action w
 
 migrate: ## Apply database migrations
 	${VENV_BIN}/python manage.py migrate --no-input
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ requirements-test: ## Install requirements for testing
 production-requirements:
 	pip install -r requirements.txt --exists-action w
 
-migrate:
+migrate: ## Apply database migrations
 	${VENV_BIN}/python manage.py migrate --no-input
 
 runserver:  ## Run django development server
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ dev.up:
 
 dev.provision:
 	docker exec -t edx.devstack.mysql /bin/bash -c 'mysql -uroot <<< "create database if not exists blockstore_db;"'
-	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && source /blockstore/venv/bin/activate && make requirements && make migrate'
+	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && make requirements && make migrate'
 
 stop:
 	docker-compose --project-name blockstore -f docker-compose.yml stop
@@ -52,7 +52,7 @@ production-requirements:
 	pip install -r requirements.txt --exists-action w
 
 migrate:
-	python manage.py migrate
+	${VENV_BIN}/python manage.py migrate --no-input
 
 runserver:  ## Run django development server
 	${VENV_BIN}/python manage.py runserver 0.0.0.0:18250
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ dev.up:
 
 dev.provision:
 	docker exec -t edx.devstack.mysql /bin/bash -c 'mysql -uroot <<< "create database if not exists blockstore_db;"'
-	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && make requirements && make migrate'
+	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && make requirements && make migrate-dev'
 
 stop:
 	docker-compose --project-name blockstore -f docker-compose.yml stop
@@ -48,7 +48,13 @@ requirements: ## Install requirements for development
 requirements-test: ## Install requirements for testing
 	${VENV_BIN}/pip install -qr requirements/test.txt --exists-action w
 
-migrate:  ## Apply database migrations
+production-requirements:
+	pip install -r requirements.txt --exists-action w
+
+migrate:
+	python manage.py migrate
+
+migrate-dev:  ## Apply database migrations
 	${VENV_BIN}/python manage.py migrate --no-input
 
 runserver:  ## Run django development server
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ dev.up:
 
 dev.provision:
 	docker exec -t edx.devstack.mysql /bin/bash -c 'mysql -uroot <<< "create database if not exists blockstore_db;"'
-	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && make requirements && make migrate-dev'
+	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && source /blockstore/venv/bin/activate && make requirements && make migrate'
 
 stop:
 	docker-compose --project-name blockstore -f docker-compose.yml stop
@@ -53,9 +53,6 @@ production-requirements:
 
 migrate:
 	python manage.py migrate
-
-migrate-dev:  ## Apply database migrations
-	${VENV_BIN}/python manage.py migrate --no-input
 
 runserver:  ## Run django development server
 	${VENV_BIN}/python manage.py runserver 0.0.0.0:18250
--- a/blockstore/settings/base.py
+++ b/blockstore/settings/base.py
@@ -205,6 +205,7 @@ LOGIN_REDIRECT_URL = '/admin/'
 PLATFORM_NAME = 'Your Platform Name Here'
 # END OPENEDX-SPECIFIC CONFIGURATION
 
+
 # Set up logging for development use (logging to stdout)
 LOGGING = {
     'version': 1,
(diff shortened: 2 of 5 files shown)


blockstore/settings/production.py
```diff
@@ -13,7 +13,7 @@
 LOGGING['handlers']['local']['level'] = 'INFO'
 
 CONFIG_FILE = get_env_setting('BLOCKSTORE_CFG')
-with open(CONFIG_FILE, encoding='utf-8') as f:
+with open(CONFIG_FILE) as f:
```
bradenmacdonald: What was fixed by removing `, encoding='utf-8'`?

requirements/production.txt
```diff
@@ -5,3 +5,5 @@ gevent
 gunicorn
 mysqlclient
 PyYAML
+python-memcached
+sqlparse
```
bradenmacdonald: Any idea why `sqlparse` is only needed in production? It seems like it should be needed anytime we run migrations. Maybe one of the other development dependencies installs it? I'm asking because it seems like we should move it to `base.txt` instead.

blockstore/settings/production.py
pkulkark: There was an exception raised while running migrations: 
`TypeError: 'encoding' is an invalid keyword argument for this function`
Removing the argument solved it. Also verified in other service repositories, like [this](https://github.com/open-craft/ecommerce/blob/master/ecommerce/settings/production.py#L47) for example, and saw that this wasn't used.

requirements/production.txt
pkulkark: Yes the `django-debug-toolbar` package included in `local.txt` installs the `sqlparse` module

Makefile
```diff
@@ -48,7 +48,13 @@ requirements: ## Install requirements for development
...
 
-migrate:  ## Apply database migrations
+production-requirements:
+	pip install -r requirements.txt --exists-action w
+
+migrate:
+	python manage.py migrate
+
+migrate-dev:  ## Apply database migrations
 	${VENV_BIN}/python manage.py migrate --no-input
```
bradenmacdonald: Does this version not work in production? I'd rather just have one single `migrate` target that works for both if possible.

requirements/production.txt
bradenmacdonald: Ok, but if we need it as well, it's better just to put it in `base.txt` rather than relying on django-debug-toolbar to install it sometimes.

Makefile
```diff
@@ -26,7 +26,7 @@ dev.up:
 
 dev.provision:
 	docker exec -t edx.devstack.mysql /bin/bash -c 'mysql -uroot <<< "create database if not exists blockstore_db;"'
-	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && make requirements && make migrate'
+	docker exec -t edx.devstack.blockstore /bin/bash -c 'source ~/.bashrc && source /blockstore/venv/bin/activate && make requirements && make migrate'
```
symbolist: The path needs to be adjustable via `VIRTUAL_ENV` so we shouldn't hard code it here. It should be possible to set it by passing it in `blockstore_environment` variable and then use ${VENV_BIN} before `/pip` and `/python`.

Makefile
```diff
@@ -48,8 +48,11 @@ requirements: ## Install requirements for development
 requirements-test: ## Install requirements for testing
 	${VENV_BIN}/pip install -qr requirements/test.txt --exists-action w
 
-migrate:  ## Apply database migrations
-	${VENV_BIN}/python manage.py migrate --no-input
+production-requirements:
+	pip install -r requirements.txt --exists-action w
+
+migrate:
+	python manage.py migrate
```
symbolist: Any specific reason we want to omit the `--no-input`?

Makefile
pkulkark: The virtual env paths are different for production and devstack. From what I understand, in production, everything should be within the `/edx/app` folder, but for the devstack we put the venv directory in `/blockstore`. Would having a similar directory structure (i.e. `/edx/app/blockstore/venvs/blockstore`) in the devstack make sense?

symbolist: Yup, that is why we just need to pass in the right `VIRTUAL_ENV` value and the rest should work. In devstack the `VIRTUAL_ENV` value is passed in from the [Dockerfile](https://github.com/open-craft/blockstore/blob/master/Dockerfile-dev#L5). In production, we can set it to wherever the venv is from Ansible. Does the make file in production not see the environment variables specified in `blockstore_environment`?

pkulkark: Makes sense. I missed to notice the `?` in front of `VIRTUAL_ENV` and assumed it was being hard-coded anyway. I've reverted the change and have added it in `blockstore_environment`.

requirements/production.txt
pkulkark: Done

Makefile
pkulkark: Done

Makefile
```diff
@@ -48,7 +48,10 @@ requirements: ## Install requirements for development
 requirements-test: ## Install requirements for testing
 	${VENV_BIN}/pip install -qr requirements/test.txt --exists-action w
 
-migrate:  ## Apply database migrations
+production-requirements:
+	pip install -r requirements.txt --exists-action w
```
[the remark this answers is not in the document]
pkulkark: Done

Makefile
pkulkark: It was an oversight. I've included it now.

blockstore/settings/base.py
```diff
@@ -222,6 +223,12 @@ def root(*x):
             'formatter': 'standard',
             'stream': 'ext://sys.stdout',
         },
+        'local': {
+            'level': 'INFO',
+            'class': 'logging.StreamHandler',
```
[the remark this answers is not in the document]
pkulkark: As per the official [docs](https://docs.python.org/2/library/logging.handlers.html), StreamHandler sends the logging output to streams (stdout or stderr) or any file-like objects. SysLogHandler is used for sending logging output to a remote or local Unix syslog.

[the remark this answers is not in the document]
symbolist: Okay. @bradenmacdonald Shouldn't we be using `SysLogHandler` as well or should we get the edX ops team opinion on this?

symbolist:
@pkulkark Besides that open question about the log handler, everything else looks 👍.

blockstore/settings/base.py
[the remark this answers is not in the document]
bradenmacdonald: @symbolist  @pkulkark  we should do whatever the other edx services are doing. Sounds like `SysLogHandler` then. It also seems that we should configure all the actual loggers below to log to both `console` and `local` - see https://github.com/edx/course-discovery/blob/1434b1b1c6cf8e3cc2525e60846610b2dc02d907/course_discovery/settings/base.py#L288-L336

After all, if `local` and `console` are both just logging to stdout, there's no point in defining them separately.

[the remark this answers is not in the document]
pkulkark: Thanks @bradenmacdonald @symbolist ! I've changed it to `SysLogHandler` based on other services and configured the loggers as required. Let me know if any other changes are required.

bradenmacdonald:
@pkulkark After this has merged, I'm seeing some new log-related errors on my blockstore devstack during every web request:

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/logging/handlers.py", line 939, in emit
    self._connect_unixsocket(self.address)
  File "/usr/local/lib/python3.6/logging/handlers.py", line 868, in _connect_unixsocket
    self.socket.connect(address)
FileNotFoundError: [Errno 2] No such file or directory
Call stack:
  File "/usr/local/lib/python3.6/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/usr/local/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.6/socketserver.py", line 651, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/local/lib/python3.6/socketserver.py", line 361, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/local/lib/python3.6/socketserver.py", line 721, in __init__
    self.handle()
  File "/blockstore/venv/lib/python3.6/site-packages/django/core/servers/basehttp.py", line 155, in handle
    handler.run(self.server.get_app())
  File "/usr/local/lib/python3.6/wsgiref/handlers.py", line 138, in run
    self.finish_response()
  File "/usr/local/lib/python3.6/wsgiref/handlers.py", line 183, in finish_response
    self.close()
  File "/usr/local/lib/python3.6/wsgiref/simple_server.py", line 35, in close
    self.status.split(' ',1)[0], self.bytes_sent
  File "/usr/local/lib/python3.6/http/server.py", line 536, in log_request
    self.requestline, str(code), str(size))
  File "/blockstore/venv/lib/python3.6/site-packages/django/core/servers/basehttp.py", line 124, in log_message
    level(format, *args, extra=extra)
Message: '"%s" %s %s'
Arguments: ('GET /static/admin/img/tooltag-add.svg HTTP/1.1', '200', '331')
```

bradenmacdonald:
^ Because `/dev/log` does not exist

pkulkark:
@bradenmacdonald Thanks for catching this! I've fixed this issue based on what other services are doing like [this](https://github.com/open-craft/ecommerce/blob/master/ecommerce/settings/devstack.py#L7-L10).
I tested the fix on my devstack raised a separate PR - #25. Let me know if anything else is required.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5474](https://openedx.atlassian.net/browse/OSPR-5474) so that we can track it in Jira.

PR #20, opened 2018-11-18
Init search engine.

symbolist:
## Description

This is the first iteration of search capabilities in Blockstore. It adds the following:

- An index and document mapping which can store the data needed for first version of LX search.
- A REST API which can be used to create, edit and delete ES documents in the default index and to query them.

- This has the minimum amount of functionality needed for the first prototype of the LX search interface. A lot more will be added once we are clear what the data sources are going to be, what the shape of the ownership, authorship, and permissions data is going to be and what query patterns will need to be supported.
- ES queries are described in JSON and ES [expects](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) queries to use a GET request with a body (it does support a query parameters format but that is very limited) . I tried to get [django-elasticsearch-dsl-drf](https://django-elasticsearch-dsl-drf.readthedocs.io/en/0.16.3/filtering_usage_examples.html) integrated as well as to come up with a query parameters format for the ES Query DSL. However, the DSL is quite complex with many capabilities and at this point it is not clear what subset we would want the REST API to be restricted to. So decided to go with POST request to `/index/<index_name>/search/` which forwards the query to ES and returns the response. At the later stage once we have more clarity about the query patterns we can look into if we can achieve all those with a GET request.
- Changing the mapping of a field in an index, requires creating a new index and re-indexing that data into it. Index aliases are used to make this work seamlessly. Support will need to be added for migrations. An option is to use [`django-elastic-migrations`](https://github.com/HBS-HBX/django-elastic-migrations). But in any case some changes will need to be made to the index and document classes.
- ES is not supposed to be [primary storage](https://www.elastic.co/guide/en/elasticsearch/resiliency/current/index.html). We will need to figure out how to re-acquire all the data that is going to be in an index in case there is a loss.

## Test Instructions

To play around with the API:

- Visit http://localhost:18250/api/v1/index/blocks/ to create, view, edit and delete documents. The DRF HTML form can be used to create and edit the data in a document. For some request examples, see `/blockstore/apps/search/tests/test_views.py`.
- Visit http://localhost:18250/api/v1/index/blocks/ and POST a query from the DRF "Raw data" form. The import query parameters are:
  - https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-query.html
  - https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html
  - https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html
  - https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-post-filter.html

Examples:

This will return documents whose `summary.title` or `summary.description` fields contain the search term sorted in descending order by the `_score` (this is a numeric value assigned to the result by ES). `from` indicates the response should contain results from the i-th result and `size` indicates the number of results to return in the request.

```
{
  "from" : 0, "size" : 5,
  "sort" : [
    { "_score" : "desc" }
  ],
  "query": {
    "bool": { 
      "should": [
        { "match": { "summary.title":   "<query>" }}, 
        { "match": { "summary.description": "<query>" }}  
      ]
    }
  }
}
``` 

To include counts for the facets we can add aggregations.

```
{
  "from" : 0, "size" : 5,
  "sort" : [
    { "_score" : "desc" }
  ],
  "query": {
    "bool": { 
      "should": [
        { "match": { "summary.title":   "<query>" }}, 
        { "match": { "summary.description": "<query>" }}  
      ]
    }
  },
 "aggs": {
    "ownership": {
      "terms": { "field": "ownership.org_id" } 
    },
    "tags": {
      "terms": { "field": "tags.paths" } 
    }
  }
}
```

The response will then contain an `aggregations` field with the data for the buckets:
```
 "aggregations": {
        "ownership": {
            "doc_count_error_upper_bound": 0,
            "sum_other_doc_count": 0,
            "buckets": [
                {
                    "key": "edX",
                    "doc_count": 3
                },
                {
                    "key": "OpenCraft",
                    "doc_count": 1
                }
            ]
        },
        "tags": {
            "doc_count_error_upper_bound": 0,
            "sum_other_doc_count": 0,
            "buckets": [
                {
                    "key": "chemistry",
                    "doc_count": 2
                },
                {
                    "key": "green",
                    "doc_count": 2
                },
                {
                    "key": "physics",
                    "doc_count": 2
                },
                {
                    "key": "astronomy",
                    "doc_count": 1
                },
                {
                    "key": "red",
                    "doc_count": 1
                }
            ]
        }
    },
```

### DIFF
--- a/blockstore/apps/search/apps.py
+++ b/blockstore/apps/search/apps.py
@@ -1,0 +1,26 @@
+""" AppConfig for search app. """
+
+from django.apps import AppConfig
+from django.conf import settings
+from elasticsearch_dsl import connections, Index
+
+from .documents import BlockDocument
+
+
+class SearchConfig(AppConfig):
+
+    name = 'blockstore.apps.search'
+    verbose_name = 'Blockstore Search'
+
+    def ready(self):
+
+        # Create the ElasticSearch connections.
+        connections.configure(**settings.ELASTICSEARCH)
+
+        index = Index('blocks')
+        index.document(BlockDocument)
+        BlockDocument._index = index  # pylint: disable=protected-access
+
+        # Create the BlockDocument index and mappings.
+        # TODO: This will need to be done in a migration.
+        index.save()
--- a/blockstore/apps/search/documents.py
+++ b/blockstore/apps/search/documents.py
@@ -1,0 +1,28 @@
+"""
+Documents for search app.
+"""
+
+from .core.documents import Document
+from .fieldsets import (
+    AnalyticsFieldSet,
+    AuthorshipFieldSet,
+    EntityFieldSet,
+    OwnernershipFieldSet,
+    SummaryFieldSet,
+    TagsFieldSet,
+)
+
+
+class BlockDocument(Document):
+    """
+    Document for entities to be indexed in BlocksIndex.
+    """
+
+    FIELD_SETS = (
+        EntityFieldSet,
+        SummaryFieldSet,
+        OwnernershipFieldSet,
+        AuthorshipFieldSet,
+        TagsFieldSet,
+        AnalyticsFieldSet,
+    )
--- a/blockstore/apps/search/fieldsets.py
+++ b/blockstore/apps/search/fieldsets.py
@@ -1,0 +1,122 @@
+"""
+FieldSets for search app.
+"""
+
+from elasticsearch_dsl import (
+    Keyword,
+    InnerDoc,
+    Integer,
+    Text,
+)
+
+from rest_framework import serializers
+
+from .core.fieldsets import FieldSet
+from .core.serializers import FieldSetSerializer
+
+
+class EntityFieldSet(FieldSet):
+    """
+    FieldSet for core data of the entity.
+    """
+
+    name = 'entity'
+
+    class Document(InnerDoc):
+
+        # Since external identifiers of different types of entities may conflict, they should not be stored in the
+        # document id field which needs to be unique. Store them as (type, id) in these instead.
+        type = Keyword()
+        id = Keyword()
+
+    class Serializer(FieldSetSerializer):
+
+        type = serializers.CharField()
+        id = serializers.CharField()
+
+
+class SummaryFieldSet(FieldSet):
+    """
+    FieldSet for summary data of the entity.
+    """
+
+    name = 'summary'
+
+    class Document(InnerDoc):
+
+        title = Text()
+        description = Text()
+        image = Text(index=False)
+
+    class Serializer(FieldSetSerializer):
+
+        title = serializers.CharField(allow_null=True, required=False)
+        description = serializers.CharField(allow_null=True, required=False)
+        image = serializers.URLField(allow_null=True, required=False)
+
+
+class OwnernershipFieldSet(FieldSet):
+    """
+    FieldSet for ownership data of the entity.
+    """
+
+    name = 'ownership'
+
+    class Document(InnerDoc):
+
+        org_id = Keyword()
+
+    class Serializer(FieldSetSerializer):
+
+        org_id = serializers.CharField(allow_null=True, required=False)
+
+
+class AuthorshipFieldSet(FieldSet):
+    """
+    FieldSet for authorship data of the entity.
+    """
+
+    name = 'authorship'
+
+    class Document(InnerDoc):
+
+        author_ids = Keyword(multi=True)
+
+    class Serializer(FieldSetSerializer):
+
+        author_ids = serializers.ListField(child=serializers.CharField(), required=False)
+
+
+class TagsFieldSet(FieldSet):
+    """
+    FieldSet for tags associated with the entity.
+    """
+    name = 'tags'
+
+    class Document(InnerDoc):
+
+        paths = Keyword(multi=True)  # Materialized path.
+
+    class Serializer(FieldSetSerializer):
+
+        paths = serializers.ListField(child=serializers.CharField(), required=False)
+
+
+class AnalyticsFieldSet(FieldSet):
+    """
+    FieldSet for analytics data.
+    """
+
+    name = 'analytics'
+
+    class Document(InnerDoc):
+
+        favorites = Integer()
+        remixes = Integer()
+        views = Integer()
+
+    class Serializer(FieldSetSerializer):
+
+        favorites = serializers.IntegerField(allow_null=True, required=False)
+        remixes = serializers.IntegerField(allow_null=True, required=False)
+        views = serializers.IntegerField(allow_null=True, required=False)
--- a/blockstore/apps/search/tests/test_documents.py
+++ b/blockstore/apps/search/tests/test_documents.py
@@ -1,0 +1,43 @@
+""" Tests for views. """
+
+from elasticsearch_dsl import Object
+
+from ..core.serializers import FieldSetSerializer
+from ..core.tests.test_base import DocumentBaseTestCase
+from ..documents import BlockDocument
+from ..fieldsets import (
+    AnalyticsFieldSet,
+    AuthorshipFieldSet,
+    EntityFieldSet,
+    OwnernershipFieldSet,
+    SummaryFieldSet,
+    TagsFieldSet,
+)
+
+
+class BlockDocumentTestCase(DocumentBaseTestCase):
+    """ Tests for BlockDocument. """
+
+    document_class = BlockDocument
+
+    def test_fieldsets_initialization(self):
+
+        for field_set_class in BlockDocument.FIELD_SETS:
+
+            document_field = BlockDocument._doc_type.mapping[field_set_class.name]  # pylint: disable=protected-access
+            self.assertIsInstance(document_field, Object)
+
+            document_serializer = BlockDocument.Serializer()
+            self.assertIsInstance(document_serializer.fields[field_set_class.name], FieldSetSerializer)
+
+    def test_document_create(self):
+
+        document = BlockDocument(
+            analytics=AnalyticsFieldSet.Document(favorites=8, remixes=2, views=200),
+            authorship=AuthorshipFieldSet.Document(author_ids=['First author', 'Second author']),
+            entity=EntityFieldSet.Document(type='xblock', id='block-v1:Blockstore+Course+type@html+block@030e35c'),
+            ownership=OwnernershipFieldSet.Document(org_id='Blockstore'),
+            summary=SummaryFieldSet.Document(title='Introduction', description='Welcome!', image='http//image.png'),
+            tags=TagsFieldSet.Document(paths=['15:easy:', '200:animal:mammal:lion:']),
+        )
+        document.save()
--- a/blockstore/apps/search/tests/test_views.py
+++ b/blockstore/apps/search/tests/test_views.py
@@ -1,0 +1,229 @@
+""" Tests for views. """
+
+from elasticsearch.exceptions import NotFoundError
+import pytest
+
+from ..core.tests.test_base import ViewsBaseTestCase
+from ..documents import BlockDocument
+from ..fieldsets import (
+    SummaryFieldSet,
+)
+
+
+class BlockDocumentsViewSetTestCase(ViewsBaseTestCase):
+    """ Tests for BlockDocumentsViewSet. """
+
+    document_class = BlockDocument
+
+    list_path = 'api:v1:index-blocks-list'
+    detail_path = 'api:v1:index-blocks-detail'
+
+    def test_list(self):
+
+        response = self.response(self.list_path)
+        self.assertEqual(response.data['results'], [])
+
+        titles = {'Video 1', 'Video 2', 'Video 3'}
+        for title in titles:
+            document = BlockDocument(summary=SummaryFieldSet.Document(title=title))
+            document.save(refresh=True)
+
+        response = self.response(self.list_path)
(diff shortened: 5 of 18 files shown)


symbolist:
@bradenmacdonald @ormsbee @pomegranited I have added the design notes in `/search/__init__.py` so that it is easy to have threaded comments. Would like your thoughts, particularly around the data model.

ormsbee:
Is the `uuid` associated with documents the Bundle UUID? If that's the case, do we have the ability for searches to index individual BundleVersions? Or is that overkill?

blockstore/apps/search/__init__.py
```diff
@@ -0,0 +1,57 @@
+"""
```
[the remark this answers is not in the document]
symbolist: Sure, will do when I update.

blockstore/apps/search/fieldsets.py
```diff
@@ -0,0 +1,119 @@
...
+class AuthorshipFieldSet(FieldSet):
+    """
+    FieldSet for authorship data of the entity.
+    """
+
+    name = 'authorship'
+
+    class Document(InnerDoc):
+
+        author_ids = Keyword(multi=True)
```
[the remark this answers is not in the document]
symbolist: For the short term we can put the names in here.

For the long term, there are a few interesting questions here. For objects (author or org in this case) which can be associated with many entities in the index, do we just want to store stable identifiers or other metadata as well with each entity? If we do, in case the metadata (author name or org name) is edited, it would require re-indexing all the entities with which that metadata was stored.

The other option is that since author and org metadata is going to be stored in Blockstore, we can fill in the values for the results which are being returned in the query.

One other consideration is the performance implications of the exact data structure we use to store the metadata. `nested` fields results in each nested object indexed as a separate document while object arrays do not; but have querying limitations.

https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

blockstore/apps/search/fieldsets.py
```diff
@@ -0,0 +1,119 @@
...
+    InnerDoc as ESInnerDoc,
+    Integer,
+    Text,
+)
+
+from rest_framework import serializers
+from .serializers import FieldSetSerializer
+
+
+class InnerDoc(ESInnerDoc):
```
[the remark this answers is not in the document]
symbolist: I was experimenting with a few defaults; will remove it if it turns out to not be needed.

blockstore/apps/search/fieldsets.py
```diff
@@ -0,0 +1,119 @@
...
+    FieldSet for summary data of the entity.
+    """
+
+    name = 'summary'
+
+    class Document(InnerDoc):
+
+        title = Text()
+        description = Text()
+        image = Text(index=False)
```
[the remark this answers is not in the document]
symbolist: To be clear, this is supposed to be a URL to the image and not the binary content and it is set to just be stored and not be searchable (`index=False` does that). Unlike authors and orgs for which we will likely have database entities, the image may be in the XML or could be a video still etc so yup the idea here is to push the image URL in the ES entity so that it can be returned along with the results. The view layer can use signed URLs if required as well.

In general it is normal to have non-searchable data saved along with the entities in ES.

symbolist:
@bradenmacdonald 

> Does this support doing a single search that could return Blocks, Units, Pathways, and Courses that match?

This initial version is storing all these entities in one index.

It is possible to have one query do a search across multiple indices, so we can store them in different indexes but our data/query patterns are such that one index is going to be sufficient. Very roughly speaking multiple indices are useful if large percentage of queries only hit subsets of data so those queries can only hit the relevant index and the broader queries can hit more of them.

It may be useful to have one for Studio and one for LX in case the later needs to have data that is not needed by the former.

@ormsbee 

> Is the uuid associated with documents the Bundle UUID?

The Document `uuid` is the id field of ElasticSearch documents. It is possible to specify it when creating the document but there are two reasons why this implementation doesn't (though it is just ):
1. I need to look into this further but the [default routing](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_automatic_id_generation
) of documents into shards is based on the id value and the auto-ids lead to a more equal distribution of documents across shards.
2. Since it can store entities of different types, there can be an id collision. The `entity_type` and `entity_id` fields are supposed to store that unique pair of values.

> If that's the case, do we have the ability for searches to index individual BundleVersions? Or is that overkill?

We can of course index individual bundleversions. To be honest, I was thinking about it more from the LX interface point of view so a couple of questions:
1. Should there be an option where the search results are grouped by Bundle? i.e. all the matching documents are raised to the position of the top result from the Bundle.
2. If we index all the versions of an entity, in search results should all the matching versions be grouped together? It will add an overhead to the regular queries because will have to filter out the non-latest version for each entity. Another option is that we can have two indices, ones which stores latest versions only and the other which stores all versions.

ormsbee:
> We can of course index individual bundleversions. To be honest, I was thinking about it more from the LX interface point of view so a couple of questions:
>
> Should there be an option where the search results are grouped by Bundle? i.e. all the matching documents are raised to the position of the top result from the Bundle.
>
> If we index all the versions of an entity, in search results should all the matching versions be grouped together? It will add an overhead to the regular queries because will have to filter out the non-latest version for each entity. Another option is that we can have two indices, ones which stores latest versions only and the other which stores all versions.

Yeah, I think that indexing the latest version is the right thing to do the vast majority of the time. There are some oddball use cases that would want to be able to search history (i.e. this copyrighted/obscene/illegal thing needs to be removed), but it's probably not worth exploding the core model to accommodate it. I like the idea of having two indices, one that stores the latest and the other that stores all, but I'm also perfectly happy to punt the one that stores all until we have stronger use case evidence than my hunch.

blockstore/apps/search/apps.py
```diff
@@ -0,0 +1,26 @@
...
+    verbose_name = 'Blockstore Search'
+
+    def ready(self):
+
+        # Create the ElasticSearch connections.
+        connections.configure(**settings.ELASTICSEARCH)
+
+        index = Index('blocks')
+        index.document(BlockDocument)
+        BlockDocument._index = index  # pylint: disable=protected-access
```
symbolist: This will get moved to a better place once we figure out how to handle migrations.

blockstore/apps/search/documents.py
```diff
@@ -0,0 +1,28 @@
...
+    TagsFieldSet,
+)
+
+
+class BlockDocument(Document):
+    """
+    Document for entities to be indexed in BlocksIndex.
+    """
+
+    FIELD_SETS = (
```
symbolist: We can figure out how to make this configurable but there are too many unknowns right now so going with hardcoding this list for now.

blockstore/apps/search/fieldsets.py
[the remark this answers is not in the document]
symbolist: Since we do not have a model for storing authors and owners (any related metadata), we can just put the whole name here for now. See tests for example.

blockstore/apps/search/fieldsets.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    FieldSet for core data of the entity.
+    """
+
+    name = 'entity'
+
+    class Document(InnerDoc):
+
+        type = Keyword()
+        id = Keyword()  # Since ids of different types of entities may conflict, we do not store them in id field.
```
bradenmacdonald: What do you mean, we do not store them in `id` field? This looks like we are storing the ID in the `id` field?

blockstore/apps/search/tests/test_documents.py
```diff
@@ -0,0 +1,43 @@
...
+
+            document_serializer = BlockDocument.Serializer()
+            assert isinstance(document_serializer.fields[field_set_class.name], FieldSetSerializer)
+
+    def test_document_create(self):
+
+        document = BlockDocument(
+            analytics=AnalyticsFieldSet.Document(favorites=8, remixes=2, views=200),
+            authorship=AuthorshipFieldSet.Document(author_ids=['First author', 'Second author']),
+            entity=EntityFieldSet.Document(type='html', id='html1'),
```
bradenmacdonald: A more realistic entity ID would be `type='xblock', id='block-v1:...'`

What exactly is in this index, btw? Is it just XBlocks, or is it XBlocks, courses, pathways, and libraries? If it's just XBlocks, we only need the UsageKey and the `type` is not part of the ID. But if we're mixing XBlocks, pathways, etc. then we might need the `type` and the `id` which together comprise the compound key.

blockstore/apps/search/tests/test_views.py
```diff
@@ -0,0 +1,229 @@
...
+
+    document_class = BlockDocument
+
+    list_path = 'api:v1:index-blocks-list'
+    detail_path = 'api:v1:index-blocks-detail'
+
+    def test_list(self):
+
+        response = self.response(self.list_path)
+        assert response.data['results'] == []
```
bradenmacdonald: I believe we shouldn't use the `assert` keyword since it can be optimized out... https://stackoverflow.com/a/6361186

blockstore/apps/search/tests/test_views.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # Add some data
+        self.response(
+            self.detail_path,
+            method='patch',
+            kwargs={'uuid': str(document.uuid)},
+            data={
+                'analytics': {'favorites': 3},
+                'authorship': {'author_ids': ['The only author']},
+                'entity': {'type': 'video', 'id': 'block-v1:uuuuiidd'},
```
bradenmacdonald: There are two ways you could be using `type` here, and we need to clarify. One is metadata, just an informative piece of information for the frontend which says "this content is a video". I think that's what we want. If that's what you mean here, that's fine.

The other is as a namespace for IDs. In that case, the `type` should not be `video` because opaque keys like `block-v1:...`, `course-v1:...`, `gblock-v1`, etc. all share the same namespace.

docker-compose.yml
```diff
@@ -2,6 +2,16 @@ version: '3'
 
 services:
 
+  elasticsearch:
+    container_name: edx.devstack.blockstore-elasticsearch
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.4.3
+    ports:
+       - "9200:9200"
+       - "9300:9300"
```
bradenmacdonald: I'm not sure that sharing these on the host as 9200/9300 is a good idea - seems like they're likely to conflict? Perhaps use ports like 8255 and 8256 which are close in number to blockstore (8250) ?

Though I guess it's fine to leave this and only change it if there's a problem.

edX doesn't expose elasticsearch to the host at all btw: https://github.com/edx/devstack/blob/db52ce74695bc3d48ed63f1f8e02d071b22d085b/docker-compose.yml#L25-L34

blockstore/apps/search/fieldsets.py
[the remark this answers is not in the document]
symbolist: Hmm. I have avoided adding any object arrays for now since they have some [peculiar behaviors](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#_how_arrays_of_objects_are_flattened) that I need to understand better. So lets go with storing the id for now.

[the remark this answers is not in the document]
symbolist: We will be transforming the ES responses in LX service anyway, to for example map the full tag paths to their labels etc so this can just be another transform which adds data.

blockstore/apps/search/fieldsets.py
symbolist: Sorry, I meant to say "Since ids of different types of entities may conflict, we do not store them in the base document id field since that needs to be unique."

blockstore/apps/search/tests/test_documents.py
symbolist: Yes, the index can contain documents for the metadata/searchable content for any kind of entity we want searchable. It can be XBlocks, courses, pathways, even pdf files, images etc (as long as the fields are correctly filled in). And yes the `(type, id)` is acting as the compound key.

symbolist: I have updated the example to `entity=EntityFieldSet.Document(type='xblock', id='block-v1:Blockstore+Course+type@html+block@030e35c'),`

blockstore/apps/search/tests/test_views.py
symbolist: Sorry for the misleading example again. This is supposed to be a namespace for IDs. I have updated this to `'type': 'xblock'`.

docker-compose.yml
symbolist: Hmm. Having them exposed is useful for interacting with the ES REST API. I'll change them to 8255, 8256.

blockstore/apps/search/tests/test_views.py
symbolist: Hmm. The pytest examples all use `assert`: https://docs.pytest.org/en/latest/. But I have updated all instances to `self.assertEqual()` for consistency.

symbolist:
@ormsbee Would you like to review as well?

ormsbee:
@symbolist: I don't have anything useful to add. Please go ahead and merge.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5473](https://openedx.atlassian.net/browse/OSPR-5473) so that we can track it in Jira.

PR #23, opened 2018-11-22
Fix missing taxonomy_uid in templatetag

SSPJ:
## Description

@bradenmacdonald I am sorry for opening another PR on this branch, but I discovered a bug. The Django admin uses the same template for both change and add. There's no `taxonomy_uid` on the add page because it hasn't been created yet, so the call to `get_tags_in_taxonomy_hierarchically_as_dict` raises an exception when it tries to cast `taxonomy_uid` to an int.

ETA: Also fixed an issue with KeyErrors caused by uppercase letters in parent tags.

## Test Instructions

1. Go to http://0.0.0.0:18250/admin/tagstore_django/taxonomy/add/ , with and without this commit

## TODOs
- [x] Squash before merging

### DIFF
--- a/tagstore/backends/django.py
+++ b/tagstore/backends/django.py
@@ -49,7 +49,7 @@ class DjangoTagstore(Tagstore):
             defaults={'path': path},
         )
         if not created:
-            if db_tag.path != path:
+            if db_tag.path.lower() != path.lower():
                 raise ValueError("That tag already exists with a different parent tag.")
         return db_tag.name
 
@@ -94,7 +94,7 @@ class DjangoTagstore(Tagstore):
         taxonomy_uid_as_int = int(taxonomy_uid)
         for tag in TagModel.objects.filter(taxonomy_id=taxonomy_uid_as_int).order_by('path'):
             node = {'name': tag.name, 'id': tag.id, 'children': []}
-            as_tuple = Tag(taxonomy_uid=taxonomy_uid_as_int, name=tag.name.lower())
+            as_tuple = Tag(taxonomy_uid=taxonomy_uid_as_int, name=tag.name)
             all_nodes[as_tuple] = node
             all_nodes[tag.parent_tag_tuple]['children'].append(node)
         return root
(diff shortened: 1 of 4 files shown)


tagstore/backends/django.py
```diff
@@ -94,7 +94,7 @@ def get_tags_in_taxonomy_hierarchically_as_dict(self, taxonomy_uid: TaxonomyId)
         taxonomy_uid_as_int = int(taxonomy_uid)
         for tag in TagModel.objects.filter(taxonomy_id=taxonomy_uid_as_int).order_by('path'):
             node = {'name': tag.name, 'id': tag.id, 'children': []}
-            as_tuple = Tag(taxonomy_uid=taxonomy_uid_as_int, name=tag.name)
+            as_tuple = Tag(taxonomy_uid=taxonomy_uid_as_int, name=tag.name.lower())
```
bradenmacdonald: This fixes the error in the admin, but not the root cause. The root cause is that `parent_tag_tuple` returns the wrong case in the tuple - it should return the correct capitalization of the tag, but because it reads `name` out of the `path` property, it's converted to lowercase.

I think the best fix is actually to change `make_path` (and the comment above the `path` field in the model) so that the materialized path is saved with the correct case of each tag instead of being in lowercase. I don't think we actually need it saved in lowercase, since we're using lowercase collation in MySQL.

SSPJ: That makes good sense. I tried changing it in `make_path` but it caused `test_case_sensitive_tags` to fail with `ValueError: Tag testing already exists with a different parent tag.`. I will poke at it some more tomorrow.

SSPJ: Okay, this might be closer to what you had in mind. I've taken `lower()` out of `make_path`, so tags are saved in their original case. The upshot of this was that in `_add_tag_to_taxonomy` it is necessary to explicitly lowercase the path when testing for tags with multiple parents. (Because '5:BioTech' != '5:biotech' although they are the same tag.)

bradenmacdonald: Thanks, looks good!

bradenmacdonald:
:+1: Thanks!

I did notice one issue; if you have time, would you mind fixing before merging? It's that if you create a new (empty) taxonomy, the "+ Add New Tag" link does not appear. It only appears if the taxonomy already contains at least one tag.

- [x] I tested this: tested as described in the django admin
- [x] I read through the code
- [x] I checked for accessibility issues: n/a
- [x] Includes documentation: n/a

SSPJ:
Thanks for the review! I've made sure the  "+ Add New Tag" link appears on empty taxonomy pages and I'll go ahead and merge this.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5471](https://openedx.atlassian.net/browse/OSPR-5471) so that we can track it in Jira.

PR #25, opened 2018-12-02
Fix the logging error in devstack

pkulkark:
## Description

This PR fixes the logging issue seen in devstack after #18 was merged, during web-requests and running tests.

## Test Instructions

To test it:
1. Start the server inside the shell, and make any request. There should not be any logging errors seen on the console.
2. From inside the shell, run the test scripts with `make validate`. There should not be any logging errors seen.

### DIFF
--- a/blockstore/settings/local.py
+++ b/blockstore/settings/local.py
@@ -13,6 +13,11 @@ CACHES = {
 }
 # END CACHE CONFIGURATION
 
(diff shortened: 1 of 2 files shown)


edx-webhook:
Although this pull request is already merged, I've created [OSPR-5469](https://openedx.atlassian.net/browse/OSPR-5469) so that we can track it in Jira.

Issue #26, opened 2018-12-07
Comparisons: Modeling Hierarchy


ormsbee:
@bradenmacdonald, @symbolist, @pdpinch:

This is a spinoff of #16, where we can dive more into hierarchy modeling specifically. I’m trying to be even handed in my analysis, but I obviously have both an imperfect understanding of the proposals on the table as well as a personal preference, so please treat this issue description like a wiki and edit it until you feel it fairly represents the tradeoffs.

The proposals on the table are:
* Bundle Per Block (BPB)
* Hierarchical One Bundle Course (HOBC)
* Flat One Bundle Course (FOBC)
* Course Bundle, Unit Bundle (CBUB)

BPB and HOBC treat hierarchy modeling as a Blockstore primitive. BPB does this by narrowing the scope of a Bundle to only cover single Blocks, and leaning on our more-or-less original formulation of BundleVersion Links to string things together. HOBC uses a combination of file system conventions and a new Folder primitive that provides a more nuanced version of Links that allows for version-relative linking. FOBC and CBUB punt hierarchy modeling responsibility entirely to systems outside Blockstore, by storing the OLX for those relationships but not having any built-in understanding of how those relationships work.

Some of the driving use cases here:
* Re-use of non-leaf blocks like Units.
* Granular course publishing (at course, chapter, sequence, unit, and possibly even individual leaf levels)
* Adaptive use cases calling for a more flexible/dynamic traversal of content.
* Remixing of larger content that lives primarily in a course (e.g. a sequence).
* Potentially restructuring existing Course Content (CCX example)

Of those, the first one is providing the strongest push to model hierarchy explicitly.

## Bundle Per Block (BPB)

BPB’s core strength is that is is both conceptually simple and highly flexible. It can represent anything the other two can and its concept of Links is more straightforward than HOBC Folders since it’s just a M:M relationship between BundleVersions. There is a bit more complexity in grouping commits together logically (four Bundles changing could be the result of one meaningful edit), but that can be accommodated by modeling those commits as first class entities of their own and having a 1:M relationship with the created BundleVersions.

[...]

PR #27, opened 2018-12-07
Bundle Granularity writeup for Flat, One-Bundle Course proposal

ormsbee:
@symbolist, @bradenmacdonald: A hopefully more fleshed out overview of what I was proposing. I'm channeling the Nimisha-angel on my shoulder and writing it up as an arch decision record, though I realize that we're just as likely to go with one of the other proposals. (I think once we do figure this out, it needs to be arch decision record number 1 for this repo.)

### DIFF
--- a/decisions/0001-bundle-granularity.rst
+++ b/decisions/0001-bundle-granularity.rst
@@ -1,0 +1,199 @@
+================
+Bundle Granularity
+================
+
+-------
+Context
+-------
+
+Blockstore provides Bundles, which can be treated as file directories with
+versioned content. This provides a lot of flexibility, but we still need to
+determine what sets of conventions (and what extra functionality) we require to
+model our courses with OLX data.
+
+One late change to our overall philosophical outlook was to radically separate
+the human readable format used in import/export from the machine-readable format
+that provides organization and reuse capability in Blockstore itself. This
+implies shifting more of the higher level OLX -> Blockstore mapping knowledge
+to be outside of Blockstore itself, and gives us opportunities to make some
+simplifications.
+
+---------------------------------------------------
+Proposal: OLX Bundles with Flat Directory Structure
+---------------------------------------------------
+
+In this proposal, a Bundle for any chunk of OLX content has top-level folders
+for every definition in the Bundle, including both leaf and container nodes. A
+problem in a problem bank may have just one such directory for the problem
+itself, or a few directories representing a Unit and multiple leaf nodes. A
+course may potentially have thousands of directories.
+
+The major focus of this approach is operational simplicity at scale. It is
+generally less featureful than other proposals and shifts more of the work to
+OLX-aware clients.
+
+An important mental shift is that this walks back from some of the file system
+analogies of the original proposal. We still have files in Bundles, and XBlock
+content is still stored in directories with their associated static assets. But
+when we make data references across XBlocks, it's by identifiers and conventions
+and not by relative file paths. It's the job of clients to transform that into
+a more friendly file-path based format for the human readable format.
+
+XBlock Directories
+==================
+
+XBlock Directories are a client convention and not a core Blockstore primitive.
+Each directory represents a single XBlock and its associated assets. The naming
+scheme is::
+
+    /{definition-key}/
+                     {tag-type}.xml
+                     static/
+
+The ``static`` directory is optional. For a concrete example::
+
+    /problem+blockstore_granularity/
+                                    problem.xml
+                                    static/
+                                           diagram.png
+
+Meanwhile, a container looks like:
+
+    /unit+blockstore_big_questions/problem.xml
+
+When containers reference their children, it's done via a new tag. So a
+particular ``unit.xml`` might look like::
+
+    <unit display_name="">
+        <!-- Normal usage -->
+        <xblock-include definition="unit+blockstore_big_questions" />
+
+        <!-- If we want to include the same content with a modifier to the usage
+             key so that we can include it multiple times with different state
+             storage. -->
+        <xblock-include definition="unit+blockstore_big_questions" usage="alternate" />
+
+        <!-- XBlock that exists in a Linked Bundle aliased to "arch_hangout_vidoes" -->
+        <xblock-include definition="arch_hangout_videos/video+blockstore_overview"/>
+    </unit>
+
+Links
+=====
+
+Link information is stored at the Snapshot level. Since the human-readable
+format is going to be a transform, we don't have to make symlink analogies to
+make the data easier to work with in authoring. Instead, Links are a mapping of
+names to Bundle Snapshots::
+
+    {
+        "links": {
+            "arch_hangout_videos": {
+                "bundle_uuid": "3fcf5f61-bc23-41ec-9452-26d12dc3b13c",
+                "snapshot": "8f7bc6e89581591fa925fefa3819d382fd793839"
+                "dependencies": [
+                    {
+                        "bundle_uuid": "fd7ee4b3-0540-406c-98b9-dd050b7ddcb2",
+                        "snapshot": "c0c0940e4b3151908b60cecd1ef5e2aa19904676"
+                    }
+                ]
+            }
+        }
+    }
+
+Since all Snapshots capture their full dependencies, they can do cycle detection
+quickly. Since Links are explicitly named, we can keep multiple references to
+different versions of the same Bundle, which might be useful for things like
+shared code libraries.
+
+The notifications aspect of Bundles would be handled by a M:M mapping of Bundles
+(not BundleVersions), with a relationship of ``(lending_bundle, borrowing_bundle,
+still_using)``, where ``still_using`` is a boolean flag that is true if the most
+recent version of the borrowing Bundle still has a link to any version of the
+lending bundle.
+
+Changeset-Level Metadata
+========================
+
+Data about who changed what and why are stored at the BundleVersion level of
+granularity. This does mean that a Changelog for a Course would cover many
+different blocks, and that finding when a particular block changed would be
+more difficult. However, it gives a good high level view of the history of the
+work as a whole (the Course) in a simple way. In the Content Library case, each
+problem would be its own Bundle, so we'd still have tracking at the individual
+problem level.
+
+OLX-Aware Metadata
+==================
+
+Data about the OLX contents of a Bundle (e.g. tagging, search indexing) would
+need to be associated to the tuple of (BundleVersion, XBlock Directory). Asset
+association also happens at this level of granularity.
+
+Publishing Transactions
+=======================
+
+When updating a course, a new Snapshot would be created by a Draft, with all the
+changes for any given set of XBlocks. Once the Snapshot successfully completes,
+a new BundleVersion of the Course would be created to point to that Snapshot. An
+interruption in the publishing process during Snapshot creation could result in
+an orphaned Snapshot that's not pointed to by a BundleVersion, but shouldn't
+result in a broken or inconsitent state.
+
+Hierarchy Representation
+========================
+
+The list of XBlock content is represented as a flat set of directories, and all
+navigational hierarchy has to be interpreted by OLX-aware clients. A student's
+path through a course is an LMS/Compositor level concern that structural OLX is
+an input to, and the storage layer of Blockstore shouldn't need to model it.
+
+Reusing Containers
+==================
+
+To re-use a Unit, you would first make a Link to the Bundle where that Unit came
+from, and then specify the Link prefix before the directory where the Unit comes
+from::
+
+    <xblock-include definition="arch_hangout_videos/video+blockstore_overview"/>
+
+This allows for arbitrary reuse at varying levels of granularity.
+
+Intentionality and Tracking Re-use
+==================================
+
+A drawback of approach is that it requires more explicit intentionality in
+course design in order for its BundleVersion dependency tracking to be
+meaningful. You could borrow a single leaf block or container block from another
+Course, but Blockstore itself would only know that the link between the
+BundleVersions existed, not the specific items that were used. This problem is
+significantly lessened in the case of Content Libraries, since each problem has
+its own Bundle there.
+
+Another way to look at it is that Blockstore's tracking of reuse is for
+update notifications, dependency checking, and licensing enforcement. Finer
+granularity measurements are better done by something more OLX-aware. For
+instance, if I'm making a CCX that uses a sequence but I hide a Unit that isn't
+relevant to my students, should that count towards my reuse of that particular
+unit? If someone Links to an entire Unit, but really they just wanted to make
+a reference to one image in one leaf of that Unit, is it Blockstore's job to
+understand the references well enough to understand that?
+
+If we start with this kind of tracking being at the core of Blockstore's data
+model, then OLX awareness and coupling will slowly work its way into the system.
+
+Performance
+===========
+
+Storing containers in a granular way makes certain kinds of concurrent
+operations simpler, like Drafts independently publishing Units and Chapters.
+However it does make lookups slower for tree traversal, and collection of data
+from many different blocks at once.
+
+Some outlier queries we should be able to accommodate:
+
+* Some courses have nearly 100 sequentials total, all of which must be
+represented in the current Studio Outline view.
+* While the median Unit has three children, outliers have been known to have
+400+ children.
+
+(Still need to fill this out)


decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+This allows for arbitrary reuse at varying levels of granularity.
+
+Intentionality and Tracking Re-use
+==================================
+
+A drawback of approach is that it requires more explicit intentionality in
+course design in order for its BundleVersion dependency tracking to be
+meaningful. You could borrow a single leaf block or container block from another
+Course, but Blockstore itself would only know that the link between the
+BundleVersions existed, not the specific items that were used. This problem is
```
bradenmacdonald: > You could borrow a single leaf block or container block from another Course, but Blockstore itself would only know that the link between the BundleVersions existed, not the specific items that were used.

I guess that's not ideal, but what specific problems would it cause?

We could reduce this by allowing Links to specify a subfolder of the upstream bundle, and only that subfolder would appear in the downstream bundle, which provide some useful data and optimizations without going as far as my proposal in making "Folders" a feature-rich entity of their own.

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+understand the references well enough to understand that?
+
+If we start with this kind of tracking being at the core of Blockstore's data
+model, then OLX awareness and coupling will slowly work its way into the system.
+
+Performance
+===========
+
+Storing containers in a granular way makes certain kinds of concurrent
+operations simpler, like Drafts independently publishing Units and Chapters.
```
bradenmacdonald: Unless we have identified a need for independent re-use of Sections and Subsections, I'd love to just roll that up into a single `outline.xml` file (which can be transformed via a diff or a JSON ruleset in each CCX course bundle [`ADD my-custom-section AFTER course-intro`, `REMOVE advanced-topic-section`] )

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+Each directory represents a single XBlock and its associated assets. The naming
+scheme is::
+
+    /{definition-key}/
+                     {tag-type}.xml
+                     static/
+
+The ``static`` directory is optional. For a concrete example::
+
+    /problem+blockstore_granularity/
```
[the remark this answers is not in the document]
ormsbee: Sure. In terms of Nit 1, I have no strong opinion on what the ID formatting should be.

On Nit 2: I'd prefer the slug was unique as well, but I'm not sure that's compatible with existing content. I mean, in practice it will be compatible for anything Studio authored, but no necessarily custom edited stuff.

decisions/0001-bundle-granularity.rst
ormsbee: I think it's mostly about controlling expectations and how people build on it over time. I'd prefer to draw a harder line and say that the core of Blockstore just says whether you're allowed to reuse things from a Course and tracks updates to said Course (or more granular Bundle), and that anything more specific than that has to be tracked and reported by something with more OLX awareness.

My concern is that if we start building even things like directories, we'll start having to eventually bring OLX parsing and awareness into the core of Blockstore. Adding a subfolder field to this proposal can't answer the question of "what does it mean to link to that Unit", since it doesn't understand how to parse the things inside that Unit to get the leaves.

Even a smart, hierarchical directory tracking system like HOBC introduces complexities and is still only an approximation of re-use tracking. CCX might re-use a directory but only an extremely limited subset of the contents. A randomized problem from a problem bank (v2, not the Content Library of today) might specify an extremely limited range of content from a very large bank. You could link in an HTML module but only because you want that one licensed picture of Anant with a chainsaw. You might want to answer the question not of how many times my thing is reused, but how many courses outside my organization use it.

I think that OLX-aware content re-use tracking is a separate system, and it's going to need to build its own rich data model to answer the questions it cares about. But if we build a barebones solution in Blockstore that's tightly coupled with low level storage and update tracking, then at every step of the way, we'll be pushed into a corner where the far cheaper path is to twist the data model a little more and add just a little feature there, optimize this access just a bit... and then we'll have the same situation we had in the LMS with Grading and continue to have with Scheduling.

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+
+Changeset-Level Metadata
+========================
+
+Data about who changed what and why are stored at the BundleVersion level of
+granularity. This does mean that a Changelog for a Course would covered many
+different blocks, and that finding when a particular block changed would be
+more difficult. However, it gives a good high level view of the history of the
+work as a whole (the Course) in a simple way. In the Content Library case, each
+problem would be its own Bundle, so we'd still have tracking at the individual
```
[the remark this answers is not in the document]
ormsbee: > This means that a course using a Problem Library with 10,000 problems still needs 10,000 Links to bundles, right?

If we're statically linking them in using the same mechanism that Content Libraries use today, that's true. I don't think we'd want that though. FWIW, the max I know of for problems in a single course is ~900.

> And how do we represent the fact that when additional bundles are added to the library, they need to be linked into the course?

Ack, that's sloppy of me. I think that there are really two kinds of Content Libraries we're talking about, and I need to 

1. The Content Library of today, where the contents are statically copied into a Sequence at publish time, should be its own Bundle, with each problem being a directory.
2. The Problem Bank of the FUTURE, where each problem is a Bundle and we'll pulling things back more dynamically. The Problem Bank as a whole would be represented as a Collection, and Courses wouldn't have a hardcoded reference to each Bundle separately.

decisions/0001-bundle-granularity.rst
bradenmacdonald: Yeah, I'm not opposed to what you're suggesting at all, in fact I'm a bit surprised that I don't see more problems with it, which is why I was asking.

What's most important for blockstore is that we always know when something has been updated upstream, but "false positives" are probably fine (e.g. you're linking to an external course to get that picture of Anant; it's critical that you get updates if that picture is modified but it's not a big problem if you get updates when other parts of the external course are also modified, especially if there's a nice course-level changelog and/or the blockstore client that's using this data can act on the notification in a smart way.)

If we forget what I said about allowing links to specify a subfolder, what do you think about blockstore maintaining a hash of not just the contents of each file but also some kind of hash at the folder level? (It could be computed in real-time simply by iterating over the metadata of all files in the bundle and hashing their hashes, with room for optimization later). That way, even though Blockstore doesn't know that you're only using the one image from one unit/folder within that course, a client which does know which folder you're using can trivially compare the folder hash of the old and new version to decide whether or not to prompt you about the update notification it just received from Blockstore.

decisions/0001-bundle-granularity.rst
ormsbee: I think I'm more hung up on the problem from a publishing point of view. Say you want to publish a chapter, and only one chapter. If the data describing that chapter is wrapped up with the rest of the outline, then you have to be able to break that up when you store it. Which is possible, but seems like you could run into weird concurrency edge cases.

I'm totally +1 on having some separate ruleset to express transforms to structure. But again, I'd like the parsing knowledge of that to live outside of Blockstore. I can easily see things like "add a node here" evolve to things like, "get me this, but remove all the survey blocks".

decisions/0001-bundle-granularity.rst
ormsbee: I think that might be a useful optimization to make, but I'd want to test it to see if it's necessary. As you point out, we already have metadata (including hashes) for all files in the Bundle. The two most common use cases I can think of are "show me all the things that changed between that snapshot and this one" and "did this one file I care about change?", which I suspect will work acceptably fast even on large courses.

But since this would be an optimized calculation over immutable data, I think it's something we can add easily add to the format down the line.

decisions/0001-bundle-granularity.rst
bradenmacdonald: We could stick to the current convention that changes to the course outline are immediate, but changes within Units or smaller can have draft-publish workflow. However, it would be nice to do better.

This is probably too complex/messy to bother with for now, but it could offer a potential solution in the future: Imagine if in order to support the CCX use case, we expect that master course bundles have an `outline.xml` and each CCX course bundle links to the master bundle and also has its `outline-changes.xml` which has the list of customizations unique to that CCX course (add/remove/re-order sections via rules). If we've already implemented all that logic, then we could use the same thing for draft-publish: In the published version of the course would sit the normal `outline.xml` and in each draft version of the course bundle (there could be many), would be an unmodified `outline.xml` plus an `outline-changes.xml` with the changes specific to that draft version of the course. Then, we get to re-use code, we don't add any functionality to Blockstore to support that use case, and we get section/subsection-level publishing.

You touch on something that's a point of concern though whether the outline data is in one file or many: what if two authors concurrently change a section by adding/removing different subsections? S3 supports atomic writes and read-after-write consistency, but it doesn't support locking, so there's no way to read-update-write while blocking other reads. Would we want to implement a basic form of write locking at the Blockstore level?

decisions/0001-bundle-granularity.rst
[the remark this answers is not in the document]
bradenmacdonald: > The Problem Bank as a whole would be represented as a Collection, and Courses wouldn't have a hardcoded reference to each Bundle separately.

But we still need to know which courses are using which libraries to facilitate update notifications. So if a Library = Collection, then that means courses would need to be able to link to Collections as well as to Bundles?

Why not just keep it simple and say 1 Course = 1 Bundle, _and_ 1 Library = 1 Bundle ?

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+
+The major focus of this approach is operational simplicity at scale. It is
+generally less featureful than other proposals and shifts more of the work to
+OLX-aware clients.
+
+An important mental shift is that this walks back from some of the file system
+analogies of the original proposal. We still have files in Bundles, and XBlock
+content is still stored in directories with their associated static assets. But
+when we make data references across XBlocks, it's by identifiers and conventions
+and not by relative file paths. It's the job of clients to transform that into
```
symbolist: 👍

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+        <!-- Normal usage -->
+        <xblock-include definition="unit+blockstore_big_questions" />
+
+        <!-- If we want to include the same content with a modifier to the usage
+             key so that we can include it multiple times with different state
+             storage. -->
+        <xblock-include definition="unit+blockstore_big_questions" usage="alternate" />
+
+        <!-- XBlock that exists in a Linked Bundle aliased to "arch_hangout_vidoes" -->
+        <xblock-include definition="arch_hangout_videos/video+blockstore_overview"/>
```
symbolist: I am assuming the definition keys in a single bundle are going to be unique. But can one context refer to two XBlocks from two different bundles which have the same definition key?

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+Link information is stored at the Snapshot level. Since the human-readable
+format is going to be a transform, we don't have to make symlink analogies to
+make the data easier to work with in authoring. Instead, Links are a mapping of
+names to Bundle Snapshots::
+
+    {
+        "links": {
+            "arch_hangout_videos": {
+                "bundle_uuid": "3fcf5f61-bc23-41ec-9452-26d12dc3b13c",
+                "snapshot": "8f7bc6e89581591fa925fefa3819d382fd793839"
```
symbolist: Is `snapshot` what is currently `version_num`?

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+                    {
+                        "bundle_uuid": "fd7ee4b3-0540-406c-98b9-dd050b7ddcb2",
+                        "snapshot": "c0c0940e4b3151908b60cecd1ef5e2aa19904676"
+                    }
+                ]
+            }
+        }
+    }
+
+Since all Snapshots capture their full dependencies, they can do cycle detection
```
symbolist: Would there be cycle detection between XBlocks in the Bundle?

decisions/0001-bundle-granularity.rst
bradenmacdonald: > can one context refer to two XBlocks from two different bundles which have the same definition key?

I don't see why not? Specifying a unique `usage` key for each would likely be required in that case though:

```
<xblock-include definition="external_bundle_1/video+blockstore_overview" usage="vid1"/>
<xblock-include definition="external_bundle_2/video+blockstore_overview" usage="vid2"/>
```

decisions/0001-bundle-granularity.rst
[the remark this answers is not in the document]
pdpinch: If you want to move towards unique slugs, that would be OK. I presume the "client" would do some validation?

decisions/0001-bundle-granularity.rst
pdpinch: I just want to confirm I understand the usage key properly. They'll need to to be unique, right? If that's the case, could we change the example to `usage="alternate1"` to emphasize this?

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+(not BundleVersions), with a relationship of ``(lending_bundle, borrowing_bundle,
+still_using)``, where ``still_using`` is a boolean flag that is true if the most
+recent version of the borrowing Bundle still has a link to any version of the
+lending bundle.
+
+Changeset-Level Metadata
+========================
+
+Data about who changed what and why are stored at the BundleVersion level of
+granularity. This does mean that a Changelog for a Course would covered many
```
pdpinch:
```suggestion
granularity. This does mean that a Changelog for a Course would cover many
```

decisions/0001-bundle-granularity.rst
```diff
@@ -0,0 +1,199 @@
...
+association also happens at this level of granularity.
+
+Publishing Transactions
+=======================
+
+When updating a course, a new Snapshot would be created by a Draft, with all the
+changes for any given set of XBlocks. Once the Snapshot successfully completes,
+a new BundleVersion of the Course would be created to point to that Snapshot. An
+interruption in the publishing process during Snapshot creation could result in
+an orphaned Snapshot that's no pointed to by a BundleVersion, but shouldn't
```
pdpinch:
```suggestion
an orphaned Snapshot that's not pointed to by a BundleVersion, but shouldn't
```

decisions/0001-bundle-granularity.rst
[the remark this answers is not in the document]
symbolist: Hmm. The later is an interesting question. Is there going to be any validation at the Blockstore level?

decisions/0001-bundle-granularity.rst
bradenmacdonald: > If they are two different XBlocks, should there not be two different ids for them?

They have different definition IDs (some combo of source bundle UUID + folder name in source bundle) and they have different usage keys here (CourseKey is same but definition and usage are each different). So in fact there are no identifiers that are the same.

In other words, `external_bundle_1/video+blockstore_overview` and `external_bundle_2/video+blockstore_overview` are different IDs, and so are "vid1" and "vid2"

Or in other words, we should always think of definition keys as including both the bundle UUID and the human-provided slug.

decisions/0001-bundle-granularity.rst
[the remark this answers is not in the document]
ormsbee: Actually, thinking on this more, leaving the type in there might make things easier for the client to get a sense of the Bundle as a whole with fewer reads...

@bradenmacdonald: What are you using in the newer opaque key types again?

decisions/0001-bundle-granularity.rst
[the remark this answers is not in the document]
ormsbee: I think it's because I have this vague sense that grouping the problems together will make less sense as content libraries grow larger, and that it will be less necessary to send those types of notifications as we shift further away from courses. If I have a course and I care that you learn about fast fourier transforms, I feel like at some point we'll cross a bridge from "I'm going to import and use these nifty FFT problems in my course" to "I'm going to let the adaptive system teach you FFT, and then let you come back to my course so that I can teach you the thing that builds on top of that."

That being said, I think that it's a safe default going in position to say 1 Course = 1 Bundle and 1 Library = 1 Bundle, so long as there is a straightforward mapping out of that -- which I think we have as long as the problems are cleanly separable from each other. I'll update the proposal to reflect that.

decisions/0001-bundle-granularity.rst
ormsbee: > We could stick to the current convention that changes to the course outline are immediate, but changes within Units or smaller can have draft-publish workflow. However, it would be nice to do better.

I really hate that we do this. As far as I can tell, the reason this exists today is because Old Mongo had technical limitations that made it difficult to avoid and then Split came along and build a layer to emulate the brokenness to be compatible.

> <snip> course outline transform idea.

Unifying CCX and draft transformation of the outline is a really interesting idea. Could you expand on the details in #26?

> You touch on something that's a point of concern though whether the outline data is in one file or many: what if two authors concurrently change a section by adding/removing different subsections? S3 supports atomic writes and read-after-write consistency, but it doesn't support locking, so there's no way to read-update-write while blocking other writes. Would we want to implement a basic form of write locking at the Blockstore level?

I'd rather avoid creating lock files if at all possible. I think that the number of concurrent authors are small enough where it's realistic to expect them to coordinate with each other, if we can do something even as dumb as exposing an activity log so they know someone else is working on it. Higher concurrency scenarios, we punt to the git course development workflow.

But we should definitely be explicit and map out the failure modes. I'll work on that this evening.

decisions/0001-bundle-granularity.rst
ormsbee: I think it should be fairly straightforward for the client to do so?

decisions/0001-bundle-granularity.rst
bradenmacdonald: > I'd rather avoid creating lock files if at all possible.

I definitely wouldn't suggest lock files. What I had in mind was obtaining a bundle-specific [MySQL advisory lock](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html) before certain types of bundle write operations, to prevent race conditions.

The main thing I was worried about is:
* Client 1 deletes `file1.olx`
* Client 2 adds `file2.olx`
* Blockstore session for client 1 opens `bundle.json` for reading
* Blockstore session for client 2 opens `bundle.json` for reading
* Blockstore session for client 1 overwrites `bundle.json` without `file1.olx`
* Blockstore session for client 2 overwrites `bundle.json` with `file2.olx` (and with `file1.olx`)

Now `bundle.json` contains a reference to `file1.olx` but it no longer exists on S3.

bradenmacdonald:
@ormsbee Can we get this merged? Are there any open questions here that are keeping us from merging?

ormsbee:
@bradenmacdonald: I think there is room for discussion on some of the details, but I think we reached consensus on the broad approach. I'll squash and merge.

symbolist:
Wohoo!

PR #28, opened 2019-01-03
Elastic Stack configuration

symbolist:
## Description

- Adds a Make command to collect static assets.
- Fixes an import bug in `settings/production.py`.
- Adds 'rest_framework.authentication.BasicAuthentication' to production. We will later need to change this to OAuth.
- Adds missing newrelic dependency.
- Adds a docker volume for the venv so that provisioning is not needed on every up.

## Test Instructions

These changes are being used to provision servers for LX stage and preview. Check that provisioning of Blockstore service completes successfully.

### DIFF
--- a/blockstore/settings/production.py
+++ b/blockstore/settings/production.py
@@ -1,4 +1,4 @@
-from os import environ
+import os
 import yaml
 
 from blockstore.settings.base import *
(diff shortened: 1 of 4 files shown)


blockstore/settings/production.py
```diff
@@ -1,4 +1,4 @@
-from os import environ
+import os
```
symbolist: `blockstore.settings.base` is using the `environ` package.

SSPJ:
:+1: This looks good to me, @symbolist

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5456](https://openedx.atlassian.net/browse/OSPR-5456) so that we can track it in Jira.

PR #30, opened 2019-01-22
[SE-660] Tag editor on pathway edit page

SSPJ:
## Description

This PR concerns the need to write tags into tagstore via the api.

This work was done in conjunction with SE-660, however as I was working on that ticket I realized that tags are not currently being read from this api at all, although there is a class in the labxchange backend to do so.

I'm submitting this PR because I don't want the work to be wasted, but the tag editor on the pathway page is using the labxchange backend to keep things consistent with the other pathway components.

**ETA** I had to pin mysqlclient to 1.3.14 because the test were failing. Cf [this issue](https://github.com/PyMySQL/mysqlclient-python/issues/306).

## Test Instructions

Switch to this branch and start blockstore. Test that simple tag lists can be added:

`curl -H "Content-Type: application/json" -d '{"tags": ["bunch", "misc", "tags"]}' http://localhost:18250/tagstore/api/v1/entities/pathway/cb292b23-f56f-468b-a3ad-3256425a5fee/tags/ | jq`

That more complex additions by taxonomy and parent can be added:

`curl -H "Content-Type: application/json" -d '{"tags": [{"taxonomy_name": "Subject Area", "tag": "animal"},{"taxonomy_uid": 1, "tag": "zebra", "parent": "animal"}]}' http://localhost:18250/tagstore/api/v1/entities/pathway/cb292b23-f56f-468b-a3ad-3256425a5fee/tags/ | jq`

That a mix of both can be added:

`curl -H "Content-Type: application/json" -d '{"tags": [{"taxonomy_name": "FreeForm", "tag": "cell"},"plasma","DNA"]}' http://localhost:18250/tagstore/api/v1/entities/pathway/cb292b23-f56f-468b-a3ad-3256425a5fee/tags/ | jq`

## TODOs
- [ ] Squash before merging


tagstore/tagstore_rest/v1/views/entities.py
```diff
@@ -62,5 +67,69 @@ def tags(self, request, entity_type=None, pk=None):
...
         return Response(serializer.data)
+
+    def update_tags(self, request, entity_type=None, pk=None):
+        '''
+        Update tags belonging to an entity.
+        '''
+        tagstore = DjangoTagstore()
+
+        entity_obj = get_object_or_404(Entity, external_id=pk, entity_type=entity_type)
+        entity = EntityId(external_id=pk, entity_type=entity_type)
```
bradenmacdonald: Nit: You could just use `entity_obj.as_tuple` in the two places `entity` is used, without needing this line to initialize it.

tagstore/tagstore_rest/v1/views/entities.py
```diff
@@ -62,5 +67,69 @@ def tags(self, request, entity_type=None, pk=None):
...
+        freeform_obj, _ = Taxonomy.objects.get_or_create(name='FreeForm')
```
bradenmacdonald: I was originally thinking that each user, or perhaps each "course"/"context" on the platform, or perhaps each organization would get its own freeform tag taxonomy, and that such things would be managed by the client (i.e. the LMS or LabXchange), not by Tagstore itself.

However, this approach of having a single FreeForm taxonomy as a first-class thing inside Tagstore makes a lot of sense and is going to help keep things simple at first. I would just ask a few things:
* `get_or_create(name='FreeForm')` will cause issues if someone ever creates another Taxonomy called "FreeForm", which could happen. We shouldn't rely on names being unique. Can we perhaps make a constant like `FREEFORM_TAXONOMY_UID = -100` and load the FreeForm taxonomy by UID? (Using a negative UID guarantees it will never conflict with manually created ones; we could also use the ID `0`).

tagstore/tagstore_rest/v1/views/entities.py
```diff
@@ -62,5 +67,69 @@ def tags(self, request, entity_type=None, pk=None):
...
+                    if not _tag:
+                        _tag = tagstore.add_tag_to_taxonomy(tag, freeform)
+                    tagstore.add_tag_to(_tag, entity)
+                except ValueError:
+                    return Response(status=500)
+
+            # a tag within a taxonomy
+            elif isinstance(tag, dict):
+                taxonomy_uid = tag.get('taxonomy_uid', None)
+                taxonomy_name = tag.get('taxonomy_name', None)
```
bradenmacdonald: I don't think we should allow selecting a taxonomy by name. Since names aren't unique, it can lead to errors.

tagstore/tagstore_rest/v1/views/entities.py
```diff
@@ -62,5 +67,69 @@ def tags(self, request, entity_type=None, pk=None):
...
+        for tag in request.data['tags']:
+            # a single free form tag
+            if isinstance(tag, str):
+                try:
+                    _tag = tagstore.get_tag_in_taxonomy(tag, freeform)
+                    if not _tag:
+                        _tag = tagstore.add_tag_to_taxonomy(tag, freeform)
+                    tagstore.add_tag_to(_tag, entity)
+                except ValueError:
+                    return Response(status=500)
```
bradenmacdonald: `Response(status=500)` will just return a blank document and an HTTP 500 code, which can make debugging hard. It's better to just let the `ValueError` exception propagate up (which will result in a 500 error response, but this time with a clear message), or to log the exception (`logger.exception(...)`) plus return a 500 response including an error message like `{"status": "error", "reason": f"Unable to add tag {tag} to taxonomy {taxonomy.name}"}`

SSPJ:
@bradenmacdonald You made a lot of good points here; I've addressed all of them. Let me know if it or [MR 81](https://gitlab.com/opencraft/client/LabXchange/labxchange-dev/merge_requests/81) needs anything else!

tagstore/settings/base.py
```diff
@@ -0,0 +1 @@
+FREEFORM_TAXONOMY_UID = -100
```
bradenmacdonald: Please add a comment explaining what this is :)

I also think that it would be fine to put this in a `constants.py` file rather than considering it a setting. (Why would someone need to change it? And if they did, couldn't that cause problems if it was already in use?)

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5455](https://openedx.atlassian.net/browse/OSPR-5455) so that we can track it in Jira.

PR #31, opened 2019-02-01
Fix version of mysqlclient until Django 2.0

pomegranited:
## Description

Prevents `mysqlclient` from updating to `>=1.4`, to maintain compatibility with Django 1.11.

As noted in the comments, we can remove this restriction after upgrading to Django 2.

cf https://github.com/PyMySQL/mysqlclient-python/issues/306

## Test Instructions

Run `make dev.provision`

## TODOs

- [ ] Squash before merging

## Reviewer

- [ ] OpenCraft TBD

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -12,6 +12,8 @@ git+https://github.com/alanjds/drf-nested-routers.git
 edx-auth-backends
 elasticsearch>=6.0.0,<7.0.0
 elasticsearch-dsl>=6.0.0,<7.0.0
-mysqlclient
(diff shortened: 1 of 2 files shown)


pomegranited:
Already done -- was working on an old master.

edx-webhook:
Although this pull request is already closed, I've created [OSPR-5454](https://openedx.atlassian.net/browse/OSPR-5454) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #33, opened 2019-03-14
Courseware Migration Plan

ormsbee:
I got rid of some of the small details to keep it short and at a high level. Please let me know what you think.

### DIFF
--- a/decisions/0002-courseware-migration-plan.rst
+++ b/decisions/0002-courseware-migration-plan.rst
@@ -6,7 +6,8 @@ Open edX Courseware Data Migration Plan
 Context
 -------
 
-
+Please note that this migration proposal is at a very early stage of iteration
+and no portion of it has yet been accepted.
 
 
 ------------------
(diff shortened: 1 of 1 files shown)


bradenmacdonald:
Looking great, @ormsbee. At a high level, I think this sounds like the right plan.

ormsbee:
@nasthagiri, @marcotuts: Can you please look at this to make sure it sounds reasonable to you?

@marcotuts: I'm handwaving a lot about a new UX for Content Libraries that we could eventually build up into a more powerful authoring experience. I'd love to get more of your thoughts there.

ormsbee:
@nasthagiri: Also, if the overall thrust of "let's leave the LMS mostly alone while we change Studio storage" sounds good, I'll move forward with creating a DEPR ticket around Old Mongo and restarting my [Old Mongo migration PR](https://github.com/edx/edx-platform/pull/17393).

decisions/0002-courseware-migration-plan.rst
```diff
@@ -0,0 +1,176 @@
...
+
+Juniper
+=======
+
+* Old Mongo: Management command to convert Old Mongo courses to Split.
+* XBlock Conversion: Video, HTML, and Capa XModules are converted to XBlock.
+* Content Libraries: Mixed operation where existing Content Libraries are
+  supported with a conversion management command. Converted Content Libraries
+  (Problem Banks) will have a new authoring experience making them easier to
+  search and manage.
```
symbolist: Will this use the existing XBlock/architecture where the xblocks in the Library are copied over into the course?

ormsbee: Yes. The way content libraries currently copy content is ugly, but it makes for a really easy interface to swap an implementation behind.

bradenmacdonald:
@ormsbee I think this plan is quite out of date and will be superseded by what comes out of the Learning Core + Modular Learning work. Is it OK with you if I close this, and other old pending PRs/issues in this repo?

PR #34, opened 2019-04-02
Remove duplicate mysql requirement

pkulkark:
## Description

This fix removes the duplicate mysql requirements in the production.txt file. The requirement for mysql is already part of base.txt and the duplication in production.txt results in the failure of `make production-requirements` command.

### DIFF
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -3,7 +3,6 @@
 
 gevent
 gunicorn
-mysqlclient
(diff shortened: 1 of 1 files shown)


edx-webhook:
Although this pull request is already merged, I've created [OSPR-5453](https://openedx.atlassian.net/browse/OSPR-5453) so that we can track it in Jira.

PR #38, opened 2019-05-19
Convert to OAuth2 login provider instead of OIDC

bradenmacdonald:
This changes our use of `auth-backends` from the deprecated OIDC provider to the OAuth2 provider, because I rebuilt my devstack and the OIDC one wasn't working.

This is working, however there are a couple issues:
* Superuser permission doesn't seem to be applied via SSO - when I login as a user who has staff and superuser permissions, the blockstore user only has staff permission
   - perhaps related to [this](https://github.com/edx/auth-backends/pull/62) but I think I had that fix?
* The OAuth2 backend by default requests the `user_id` scope, and based on [these](https://github.com/edx/auth-backends/pull/59) [two](https://github.com/edx/edx-platform/pull/20057) PRs, that should work, but the default devstack LMS is giving me an `invalid_scope` error when `user_id` is included. Not sure why.

Test instructions: See the instructions added to the README.


bradenmacdonald:
Thanks @ormsbee!

bradenmacdonald:
@robrap Would you happen to know the default devstack LMS is giving an `invalid_scope` error when our IDA tries to use the OAuth2 SSO flow and requests the `user_id` scope from the LMS? (i.e. why `edx-auth-backends` does not work out of the box, unless I override it to not request that scope?)

I'm guessing it relates to https://github.com/edx/auth-backends/pull/59 and https://github.com/edx/edx-platform/pull/20057 but I definitely have both of those PRs merged in the branches I'm using for edx-auth-backends and for the LMS.

robrap:
I forwarded an email that should help. Let me know if you still have problems.

robrap:
FYI: @nasthagiri: another instance of trouble for someone unless and until we swap the default scopes.

bradenmacdonald:
Ah, thanks so much @robrap :) That was the missing info I needed. Everything is working well now.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5450](https://openedx.atlassian.net/browse/OSPR-5450) so that we can track it in Jira.

PR #39, opened 2019-06-12
Make compatible with Python 3.5.

symbolist:


blockstore/apps/bundles/store.py
```diff
@@ -647,7 +648,7 @@ def is_safe_file_path(path):
 
 def create_hash(start_data=b''):
     """Whenever there's a need for hashing, we use 20-byte BLAKE2b."""
-    return hashlib.blake2b(start_data, digest_size=20)  # pylint: disable=no-member
```
symbolist: Is not available in Python 3.5.

symbolist:
@ormsbee I have made all the necessary changes if you would like to take a look (otherwise will have someone from LX review tomorrow). Still investigating why quality is failing on CircleCI.

ormsbee:
The changes look fine to me, but it looks like quality is failing to run altogether now?

symbolist:
@ormsbee Had to pin `astroid<1.6` to get it to run. Ran the job twice and it worked both times.

symbolist:
FYI @bradenmacdonald since you are using some of the type stuff in 3.6. :)

bradenmacdonald:
Hmm, that kinda sucks that we can't use f-strings and some of these other things. But thanks for making this change @symbolist.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5449](https://openedx.atlassian.net/browse/OSPR-5449) so that we can track it in Jira.

Issue #19, opened 2018-11-17
Import/Export Mapping Logic and a Dedicated CLI Client?


ormsbee:
@pdpinch, @JAAkana, @bradenmacdonald, @symbolist, @pomegranited:

Extracted out of issue #16: What is the ideal course import/export format, keeping in mind that we will have the ability to associate assets with individual leaf XBlocks.

We know that a number of technically advanced course teams generate OLX from existing banks of problems or do all their authoring in XML and use import/export as their primary publishing workflow. However, there are some other use cases we also need to keep in mind...

## Other Known Use Cases

A list of known import/export use cases compiled by @JAAkana:

### Rerunning (v1)

A course team created an empty Studio ‘shell’ for their rerun months ago, and they’re ready to finally load the content of the current run into that shell today (they’d copy over their current course automatically if they could).

### Rerunning (v2)

Course team may have originally created a rerun via auto-copying their current run, but some other version of their course turned out to be better.

### Backups

I’m going to do something risky to my course and I need a backup copy - the import needs to be exactly the same to preserve location-ids / access to student data

### Course division / course chimera (rare)

I’ve run one giant 18-week course, and I’d like to split it into 3 small courses without copy/pasting everything. Or, I’ve run 3 small courses and I want to repackage them as 2 new courses.

### Moving content between multiple instances of the platform

Most commonly, this involves moving courses from test runs on Edge to MOOC runs on edX.org. Some course teams also move content from their own instances to edX controlled ones and vice versa.

### Libraries

Teams want to use their MOOC problem banks on campus or vice-versa

### XML editing outside of Studio

Conditional modules, changing a course’s wiki slug, adding user-readable unit URLs, etc. These are sometimes edits that either cannot be supported by the Studio UX at all (because no appropriate handler was written for them), or else are really cumbersome to do with existing editors (e.g. course-wide search and replace).

### Retrieving files (rare)

I’ve uploaded a bunch of assets to Studio months ago and now I want copies of them -it is easier to get them via export than by clicking on each.

### Retrieving data (rare)

I want a bunch of info from my course that’s hard to find click-wise, so I’ll export it and use the xml: e.g. ‘what are all the youtube ids? Where did I say ‘week’ instead of ‘lesson’ as I update to convert to self-paced’?

### Seeding a new course (rare)

I have a introductory sequence that I’d like to appear in a lot of my courses - I’ll import this content and then build the rest of my course.

### Translating from Another File Format

[...]

PR #40, opened 2019-06-18
Links

ormsbee:
Adds a 'links' param to drafts and snapshots that acts very much like the existing `files` one. Deletes are currently broken. And I need to rebase.


bradenmacdonald:
Will there be an API that returns all the files in a bundle, including all the linked files? As an API consumer, I'd rather not have to manually fetch the files list from every linked bundle when I want to do something like "get all the OLX files in the bundle (and its linked bundles)"

Will link information be persisted in the database at all? If not, could race conditions create cyclic links?

ormsbee:
Still catching up from coming back from my vacation...

> Will there be an API that returns all the files in a bundle, including all the linked files?

A GET on the BundleVersion returns all files owned by that BundleVersion. It will also return the links to other BundleVersions, but won't return all their linked files -- a client would have to hit each one in turn, but that shouldn't be a major burden relative to the number of requests that client has to make to get all the files in a BundleVersion anyway.

I don't think we can make a performant API that includes all linked files unless we're willing to sharply cap either the number of Links or the number of files in a given Bundle, as the product of those can potentially get very large. Since we're modeling entire courses as a single Bundle, we're already talking thousands of files. If we want to support the other end of the spectrum (e.g. 1K Links across small Bundles), it starts to get prohibitively large to support that kind of API.

> Will link information be persisted in the database at all? If not, could race conditions create cyclic links?

Race conditions shouldn't create link cycles. Summary files are written in atomically, and each summary file has all the information it needs to have in order to verify that no cycles exist (all Links include their extended Dependencies, so we can quickly check that). We definitely can have race conditions, but it'd be a race between two sets of things that are both internally consistent, and only one would win (there'd be no merging). Database information about Links would be limited to coarse Bundle-Bundle Link relations instead of BundleVersion-BundleVersion Links, for notification purposes.

bradenmacdonald:
@ormsbee 

> We definitely can have race conditions, but it'd be a race between two sets of things that are both internally consistent, and only one would win (there'd be no merging).

Sorry, I think I'm missing something. What prevents this scenario?

| Blockstore Process 1 | Blockstore Process 2 |
|-----------|-----------|
| Want to create link from A (consumer) to B (provider) |  Want to create link from B (consumer) to A (provider)     |
| Read bundle B's summary file: no links to A |  |
|         | Read bundle A's summary file: no links to B |
| Write bundle A's summary file: add link to B | |
| | Write bundle B's summary file: add link to A |

Result: cyclic link, and the summary files contain incomplete information about the extended links?

ormsbee:
Because a Link isn't created between Bundles, it's created between existing, immutable BundleVersions. So the scenario you pointed out above could happen, but it would look like:

Published BundleVersions A1, B1 -- no Links exist in either of these.

A2 is created with a Link to B1
B2 is created with a Link to A1

So at this point, you could say there's a circular dependency between A and B, and trying to update either Link to the latest version (e.g. creating an A3 with link to B2) will fail because a cycle is detected. But nether A2 -> B1 nor B2 -> A1 represents a cycle in and of itself. Does that make sense?

bradenmacdonald:
@ormsbee Thanks, that's what I was missing :)

ormsbee:
@bradenmacdonald, @symbolist: Ready for review.

blockstore/apps/bundles/links.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Links are named because it's totally fine to have Links that point to
+    different versions of a Bundle (e.g Bundle Av1, Bundle Av2), but we want
+    some stable way of identifying and differentiating between those references.
+
+    The indirect dependencies are a flat list. We don't try to keep track of all
+    the nested dependency relations. Instead, we want the Link to answer the
+    question of "If I wanted to download this entire Bundle and all the things
+    it links to, what is the complete list of other BundleVersions that I have
+    to download to get all the dependencies?
```
bradenmacdonald: Will this be a problem for OLX parsing clients?

i.e. if we end up with a bundle that has an OLX file which says:

```
<xblock-include definition="linked_bundle1/dep2/dep3/dep4/dep5/dep6/dep7/html/intro" usage="blah" />
```
(hopefully that is an extremely unusual example, or perhaps even forbidden; but nonetheless: )

Would the client then be expected to make 7+ API calls to load the link information for those bundles one at a time in order to resolve the name of each component in the "path" to a BundleVersion, to be able to identify which BundleVersion contains the referenced file?

That may be a reasonable trade-off, just want to clarify.

blockstore/apps/bundles/tests/test_store.py
```diff
@@ -10,7 +10,11 @@
 from django.core.files.base import ContentFile
 
 from blockstore.apps.bundles.tests.storage_utils import isolate_test_storage
-from ..store import create_hash, DraftRepo, FileInfo, Snapshot, SnapshotRepo, StagedDraft
+from ..store import (
+    create_hash, DraftRepo, FileInfo,
+    Dependency, Link, LinkCollection,
+    Snapshot, SnapshotRepo, StagedDraft,
```
bradenmacdonald: I think the preferred style is just one import name per line?

blockstore/apps/api/v1/tests/test_contract.py
```diff
@@ -390,25 +385,87 @@ def test_editing_draft(self):
...
+        new_bv_data = response_data(self.client.get(new_bv_url))
+        bv_link_data = new_bv_data['snapshot']['links']['dog_library']
+
+        # The link data should be identical except that only Drafts have a
+        # 'modified' key...
+        draft_link_data.pop('modified')
+        assert draft_link_data == bv_link_data
+
+        self.client.patch(
+            course_draft_url, data={'links': {'dog_library': None}}, format='json'
```
bradenmacdonald: Shouldn't there be some code after this to check that the specified link was deleted?

ormsbee:
Working through revisions today/tonight...

> @ormsbee It seems to work as we discussed (though a few more tests for different cases would be helpful). 👍

Yeah, I got burnt out on how many iterations I had taken already that I chucked it out. I'll add more tests.

> 1. I am assuming the users of the API will have to take care of pruning links (in case they are not needed) in new newer versions?

Yes.

> 2. Can there be any performance related concerns about the depth of the dependency tree or does inlining the dependencies takes care of those?

I think that inlining dependencies takes care of most of these. The big things we want to make fast are cycle detection and finding the list of all the BundleVersions for download queuing purposes, and since we encode all transitive dependencies, this should be fast. I think the other end of it is that we have to cap the total number of dependencies -- i.e. it's an error for a BundleVersion to have more than 1K extended dependencies.

> 3. I am curious how we are planning to refer to the links in bundle content? Is that going to be a runtime concern only?

Yeah, I'm thinking runtime concern. Links have names, but what those names mean and how they're generated would be the client's responsibility, as the one making the Links in the first place. I think that flows naturally with the smart-client, simple-core vision decided in #19.

blockstore/apps/bundles/links.py
ormsbee: This is a client decision, but I don't the Studio client here should allow direct linking to things that are dependencies of dependencies. If I have a piece of content that I depend on explicitly, then I should add that as a named Link to my Bundle, and make that reference directly. Does that seem reasonable?

bradenmacdonald: Yep, that seems reasonable.

bradenmacdonald:
@ormsbee Do you have in mind a standardized way to represent a linked file within a bundle, such as via a special path? 

For example in some earlier sketches we'd talked about paths like `_links/name/path/file.xml` representing files in linked bundles.

I ask because [`BundleDefinitionLocator`](https://github.com/edx/edx-platform/blob/d58b1c2dd180f6f26947d685c9b208f07a559bc4/openedx/core/djangoapps/xblock/learning_context/keys.py#L34) must encode all the data required to locate an OLX file in blockstore, and it currently stores only `bundle_uuid` + `olx_path`. If the link information cannot or should not be encoded into the path, I'll have to add an additional `link` field (which gets complicated if it must be able to encode a link to a link?).

bradenmacdonald:
Whoops, forgot to change the merge target of my PR and accidentally merged it into here; fixed now.

ormsbee:
> @ormsbee Do you have in mind a standardized way to represent a linked file within a bundle, such as via a special path?
>
> For example in some earlier sketches we'd talked about paths like _links/name/path/file.xml representing files in linked bundles.
> 
> I ask because BundleDefinitionLocator must encode all the data required to locate an OLX file in blockstore, and it currently stores only bundle_uuid + olx_path. If the link information cannot or should not be encoded into the path, I'll have to add an additional link field (which gets complicated if it must be able to encode a link to a link?).

Sorry for not replying to this earlier. I think you've already figured this out, but just in case...

In keeping with our previous discussions and @symbolist's excellent point about dropping the file system analogy and making clients deal with things, I didn't have any special plan to map virtual paths to links. My main concern there would be collisions with actual file paths.

ormsbee:
The one thing I will say is that links are set up with one useful level of indirection in that they're named. So the intention is for things inside the Bundle to use the uniquely named Links, with the idea that the name stays stable while the thing being pointed to changes underneath. In theory we could change the Link to an entirely different Bundle.

bradenmacdonald:
@ormsbee Thanks. I re-read all these discussion later and figured that out. I've realized that in order to support links, I need to change `BundleDefinitionLocator` to point to a BundleVersion+path rather than a bundle+path, so I'm making that change.

ormsbee:
@bradenmacdonald: Link cycle fixed, added contract test for extended cycles.

PR #42, opened 2019-06-25
Add authorization to Blockstore API [SE-1255]

bradenmacdonald:
## Description

We want to ensure that even if the Blockstore REST API is accidentally exposed to the public internet, it will not accept any requests from unauthorized users/applications.

This PR restricts API access to:
* Users with superuser permission (so LMS admins can login via SSO and then use both the API and the django admin)
* Application service users that [authenticate with a valid token header](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) - this is how the LMS will authenticate itself.

Most Open edX IDAs use OAuth2 authentication; I tried using that initially but it turns out that django-oauth-toolkit is incompatible with the `utf8mb4` database encoding that Blockstore uses. Using a token header is simpler, faster, and perfectly secure, provided that the application is always served over HTTPS.

## Test Instructions

Login as a regular user then as a superuser and try to access http://localhost:18250/api/v1/

### DIFF
--- a/blockstore/apps/api/permissions.py
+++ b/blockstore/apps/api/permissions.py
@@ -15,7 +15,7 @@ class IsSuperUserOrAuthorizedApplication(permissions.BasePermission):
       service user account. Tokens can be created at /admin/authtoken/token/
 
     Blockstore does not handle detailed (object-level/bundle-level)
-    authorization because it can rely on many details (enrollment, due dates,
+    authorization because it relies on many details (enrollment, due dates,
     cohorts, etc.) that Blockstore is not aware of. Instead, applications like
     the Open edX LMS have full access to Blockstore and are responsible for
     providing a limited subset of Blockstore functionality to authorized users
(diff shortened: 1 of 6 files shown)


blockstore/apps/api/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+    DRF Permissions class that prevents any usage of the API unless
+    * The user is authenticated as a django superuser (is_superuser=True)
+    or
+    * The user is authenticated as a service user account with a valid
+      rest_framework.authtoken.models.Token associated. This is meant to be used
+      by other applications like the Open edX LMS which use this API via a
+      service user account. Tokens can be created at /admin/authtoken/token/
+
+    Blockstore does not handle detailed (object-level/bundle-level)
+    authorization because it can rely on many details (enrollment, due dates,
```
symbolist: "it relies on"

bradenmacdonald:
Thanks @symbolist!

@feanil would you be able to take a look at this and let me know if this is OK? This is adding some base authentication to blockstore's APIs as we mentioned on our last call together, since it currently relies solely on not being exposed to the public internet.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5447](https://openedx.atlassian.net/browse/OSPR-5447) so that we can track it in Jira.

PR #49, opened 2019-08-10
Allow deleting draft via REST API

bradenmacdonald:
## Description

Due to a minor oversight, it wasn't possible to delete a draft using the REST API.

## Test Instructions

Go to http://localhost:18250/api/v1/drafts and click on any draft. Notice the "Delete" button is missing (and if it was there it would raise an error). Check out this branch. Refresh. Now it's present and will work.


edx-webhook:
Although this pull request is already merged, I've created [OSPR-5429](https://openedx.atlassian.net/browse/OSPR-5429) so that we can track it in Jira.

PR #50, opened 2019-08-13
Make command to run an isolated, ephemeral blockstore server for tests and development

bradenmacdonald:
## Description

This adds a new make command, `make testserver` which you can run on your host to spin up a temporary, isolated instance of blockstore that runs at http://localhost:18251 . The idea is that this isolated instance can be used by edx-platform for automated tests of functionality that depends on Blockstore.

The tests should be written so that they do not assume any cleanup is done automatically; each test (class) should create bundles/collections when needed without ever using fixed UUIDs. That way, for example, a single Blockstore server can be used as a target for all test runs on Jenkins, a single instance of this `make testserver` command can be used for many separate local runs of the edxapp test suite during development, and tests can be parallelized.

Simply hit CTRL-C when the server is no longer useful (you're done running tests) and its container, MySQL database, and blockstore data (media) will be destroyed. Only the virtualenv volume is kept, to make subsequent runs of `make testserver` much faster.

I have set the target to be `ormsbee/links` temporarily because I need to use this to write tests that may depend on the links functionality, but I don't want it to appear in the diffs.

I'm thinking that for now, any test we write in edxapp that depends on blockstore can be feature-flagged and off by default? Eventually we'll want them running all the time, but to do that we need a version of this container that's always running on devstack and jenkins (or at least is spun up while the test suite runs).

The Dockerfiles and docker-compose files here are _much_ faster and easier to work with than the ones in `edx/configuration` and `edx/devstack`, but I assume we'll have to transition to the latter - see https://github.com/edx/configuration/pull/5352 for a start

## Test Instructions

Optional: Run a normal instance of blockstore (`make dev.up`, `make blockstore-shell`, `make runserver`)

Run `make testserver` on the host computer.

You should now have an isolated instance of blockstore. For debugging convenience, you can log in using `staff` / `edx` at http://localhost:18251/admin/ . Confirm this instance is using a separate database, and that an auth token was created at http://localhost:18251/admin/authtoken/token/ for edx-platform to use.


blockstore/apps/api/v1/tests/test_contract.py
```diff
@@ -384,6 +384,30 @@ def test_editing_draft(self):
         )
         assert korea_text == "안녕하세요!"
 
+    def test_deleting_draft(self):
```
bradenmacdonald: This is unrelated to this PR but is temporarily included because I need it for tests. Once #40 and #49 merge, I'll rebase this PR so this unrelated diff is gone.

Makefile
```diff
@@ -66,6 +66,22 @@ test: clean ## Run tests and generate coverage report
...
+	docker-compose --project-name blockstore-testserver -f docker-compose-testserver.yml exec blockstore /blockstore/venv/bin/python /blockstore/app/manage.py runserver 0.0.0.0:18251
+	# And destroy everything except the virtualenv volume (which we want to reuse to save time):
+	docker-compose --project-name blockstore-testserver -f docker-compose-testserver.yml down
+	docker exec -t edx.devstack.mysql /bin/bash -c 'mysql -uroot <<< "drop database blockstore_test_db;"'
+
+testserver-data:  # Gets run inside the testserver container
+	# Create a staff user so developers can log in and debug issues:
+	echo "from django.contrib.auth import get_user_model; User = get_user_model(); u, _ = User.objects.get_or_create(username='staff', is_staff=True, is_superuser=True); u.set_password('edx'); u.save();" | /blockstore/app/manage.py shell
+	# Configure the API auth header:
+	echo "from django.contrib.auth import get_user_model; User = get_user_model(); u, _ = User.objects.get_or_create(username='edxapp'); from rest_framework.authtoken.models import Token; Token.objects.get_or_create(user=u, key='edxapp-test-key')" | /blockstore/app/manage.py shell
```
ormsbee: Nit: Putting Python code in a Python file somewhere and cat/exec'ing it somehow from the file name might make this easier to modify and debug issues later.

bradenmacdonald: Yeah, good call. Done in 884b9bf2472f4e6848c6b06d68fdd15cfbc4d8bf

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5428](https://openedx.atlassian.net/browse/OSPR-5428) so that we can track it in Jira.

PR #51, opened 2019-08-13
Allow description to be blank when creating a bundle

bradenmacdonald:
## Description

Allow description to be blank when creating a bundle

I assume we don't care if bundles have blank descriptions? I left slug as required.

If we want to insist on requiring descriptions, it's fine to close this PR.


bradenmacdonald:
> I suspect we should just kill slug.

Good point. I haven't encountered a use case for it yet.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5427](https://openedx.atlassian.net/browse/OSPR-5427) so that we can track it in Jira.

PR #52, opened 2019-08-16
Fix: the API should only return absolute URLs

bradenmacdonald:
## Description

APIs returned by Blockstore [should](https://www.django-rest-framework.org/api-guide/reverse/) always be absolute, but currently the BundleVersion serializer sometimes returns relative URLs (if using local file media storage). This requires the client to be able to convert those to absolute URLs which is a pain.


edx-webhook:
Although this pull request is already merged, I've created [OSPR-5426](https://openedx.atlassian.net/browse/OSPR-5426) so that we can track it in Jira.

PR #53, opened 2019-08-20
Update bundle format documentation

bradenmacdonald:
## Description

This updates the bundle granularity doc to describe what I've found to work well after implementing things in https://github.com/edx/edx-platform/pull/20645

I also removed some older doc files which were conflicting or unused.


decisions/0001-bundle-granularity.rst
```diff
@@ -46,56 +46,145 @@ XBlock Directories are a client convention and not a core Blockstore primitive.
...
+when including an XBlock definition from a linked bundle. This makes the
+implementation of content libraries considerably simpler.
+
+------------------------
+Proposal: Course Bundles
+------------------------
+
+The format of course bundles has not yet been finalized but will likely be
+an OLX bundle that includes some sort of additional "outline" file, such as
+``course-outline.json`` which specifies how the various XBlock definitions it
```
ormsbee: FWIW, I think that we're going to need multiple files if we're going to support sub-section and section level publishing, but I'm happy to punt that discussion to later and leave this paragraph as-is.

bradenmacdonald:
Ah, I just deleted that whole folder. I can restore the Makefile and .py files, and maybe index.rst

decisions/0001-bundle-granularity.rst
```diff
@@ -154,7 +243,7 @@ To re-use a Unit, you would first make a Link to the Bundle where that Unit came
 from, and then specify the Link prefix before the directory where the Unit comes
 from::
 
-    <xblock-include definition="arch_hangout_videos/video+blockstore_overview"/>
+    <xblock-include definition="arch_hangout_videos/video/blockstore_overview"/>
```
bradenmacdonald: Note that another option we could use is

```
<xblock-include source="arch_hangout_videos" definition="video/blockstore_overview" />
```

which I think is more future-proof. But I'm fine with either, and went with what's in the doc here for now.

ormsbee: Yeah, I like your version better, explicitly splitting out the source Link name.

bradenmacdonald:
Updated to restore the Makefile and change the `<xblock-include` to be more explicit.

decisions/0001-bundle-granularity.rst
```diff
@@ -46,56 +46,145 @@ XBlock Directories are a client convention and not a core Blockstore primitive.
...
+
+As with OLX Bundles in general, the conventions defined here are not implemented
+nor enforced by Blockstore in any way, however they are documented here for the
+sake of having a central reference.
+
+A content library bundle is an OLX bundle that holds a collection of one or more
+XBlocks. Each XBlock definition (OLX ``definition.xml`` file) in the content
+library is either a "top-level block" or is a child of exactly one other XBlock
+definition in the same bundle. A content library bundle does not allow a single
+definition to have multiple different usages (multiple <xblock-include />
```
symbolist: Curios what the specific reason for this is?

bradenmacdonald: It keeps things simpler when there's 1:1 usage:definition within the content library. For example, when deleting a usage, if multiple usages are allowed, you must scan through the whole bundle to see if the definition is used anywhere else in order to know if the underlying OLX file should be deleted or left in the bundle. With this restriction, there's a much simpler process: deleting a usage deletes the definition too, unless the definition is from a linked bundle.

That said, with some of the latest refactoring that I did, it's probably not too hard to relax this restriction. We can always consider relaxing it in the future.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5425](https://openedx.atlassian.net/browse/OSPR-5425) so that we can track it in Jira.

PR #55, opened 2019-09-02
Add an even easier way to run the service

bradenmacdonald:
This PR reduces the `make dev.up`, [`make provision`,] `make blockstore-shell`, `make runserver`, and `make stop` steps into a single `make easyserver` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,34 +7,30 @@ help:
 	@echo ""
 	@echo "These should be run from outside the container:"
 	@echo ""
-	@echo "  dev.up                    Start Blockstore container"
-	@echo "  dev.provision             Provision Blockstore service"
-	@echo "  stop                      Stop Blockstore container"
-	@echo "  destroy                   Remove Blockstore container, network and volumes. Destructive."
-	@echo "  blockstore-shell          Run a shell on Blockstore container."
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -66,7 +62,13 @@ test: clean ## Run tests and generate coverage report
 	${VENV_BIN}/coverage xml
 	${VENV_BIN}/diff-cover coverage.xml --html-report diff-cover.html
 
-testserver:  ## Run an isolated ephemeral instance of Blockstore for use by edx-platform tests
+easyserver: dev.up dev.provision  # Start and provision a Blockstore container and run the server until CTRL-C, then stop it
+	# Now run blockstore until the user hits CTRL-C:
+	docker-compose --project-name blockstore -f docker-compose.yml exec blockstore /blockstore/venv/bin/python /blockstore/app/manage.py runserver 0.0.0.0:18250
```
symbolist: Interesting, would be useful for the other devstack as well if could pass arbitrary commands to be run inside the container.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5424](https://openedx.atlassian.net/browse/OSPR-5424) so that we can track it in Jira.

PR #56, opened 2019-09-07
Fix: could not set unicode slugs via the REST API

bradenmacdonald:
## Description

The `slug` field has `allow_unicode=True`, but the DRF serializer was rejecting unicode characters if you tried to use the REST API to create a bundle with a unicode slug.

I know we may just remove the slug field altogether but this is an easy fix for now.


edx-webhook:
Although this pull request is already merged, I've created [OSPR-5423](https://openedx.atlassian.net/browse/OSPR-5423) so that we can track it in Jira.

PR #73, opened 2020-07-17
WIP: Add new dockerfile and github action to publish it

katenels:
## Description

This will allow us to publish Docker images to Dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+# Dockerfile
+
+FROM python:3.5.7-alpine3.9
+
+ENV VIRTUAL_ENV=/blockstore/venv
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Dockerfile
```
adzuci, line 1: Thanks for putting this PR up @katenels!  I created http://dockerhub.com/openedx/blockstore so I believe it should be good to merge if the Dockerhub credentials are available.

A good next step would be to revamp this docker file so that it looks more like what we run in production.  https://github.com/edx/enterprise-catalog/blob/master/Dockerfile may be a good reference. cc @kdmccormick

PR #82, opened 2020-09-10
Upgrade blockstore to use Python 3.8 and Django 2.2+.

Kelketek:
## Description

This pull request upgrades Blockstore to use Python 3.8. It also upgrades its Django version to match that of the LMS, since the Django version we had does not officially support 3.8 and I found this to have a handful of strange side effects.

## Author Comments, Concerns, and Open Questions

I originally changed the docker-compose files to no longer use a cached volume for the venv, but after finding that it did, in fact, have pretty serious consequences on startup that it didn't seem to have when I checked initially, I decided to just change the volume name so that a new volume will be created with a new copy of Python.

I'd considered updating the docker file to just make installing those packages as part of the build, but with as many other things as this PR was touching, I felt a lighter change was a better idea.

This PR replaces https://github.com/edx/blockstore/pull/77 and https://github.com/edx/blockstore/pull/79, both of which have partial solutions. I've also added in some text to the readme to make it easier for newly onboarded devs to figure out how to run tests.

## Test Instructions

These are now in the Readme.

## Reviews
- [ ] @bradenmacdonald

PR #84, opened 2020-09-16
Revert "Upgrade blockstore to use Python 3.8 and Django 2.2+. (#82)"

kdmccormick:
```
This reverts commit b269ba903933fdf96976f5054ee12d40147e807d.

We are reverting for two reasons:
1. We're diagnosing some Blockstore deployment issues that arose in May,
   so we'd like to remove this variable for now.
2. We want Blockstore to support Python 3.5 and Python 3.8
   simultaneously for a period of time so that we can toggle
   between 3.5/3.8 in edx/configuration easily.
```

Review: @bradenmacdonald 
FYI @Kelketek


bradenmacdonald (approved):
Revert seems straightforward  :+1:

PR #85, opened 2020-09-16
Remove unused waffle references

stvstnfrd:
This passes tests locally, but I'm not immediately clear how to E2E test this locally (blockstore is new to me).

### DIFF
--- a/blockstore/settings/base.py
+++ b/blockstore/settings/base.py
@@ -44,7 +44,6 @@ THIRD_PARTY_APPS = (
     'rest_framework_swagger',
     'django_filters',
     'social_django',  # To let admin users log in using their LMS user account
-    'waffle',
     'corsheaders',
     'release_util',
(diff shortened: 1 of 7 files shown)


kdmccormick (approved):

bradenmacdonald:
> I'm not immediately clear how to E2E test this locally (blockstore is new to me).

The test process described at https://github.com/edx/blockstore#running-integration-tests should be sufficient

bradenmacdonald:
Going to close this for now:
* `django-waffle` is installed in any case because of another dependency, `edx-django-utils`
* Most instances will be using blockstore as an app, in which case the platform determines that waffle is already installed and enabled. 

So the only benefit from doing this removal now would be to disable the `waffle.middleware.WaffleMiddleware` in the independently deployed version of Blockstore, which basically only edx.org is using.

PR #104, opened 2021-05-21
Fix and re-run requirements upgrade

kdmccormick:
In this PR, we:
* modify requirements constraints to fix the requirements upgrade job,
* upgrade requirements pins, and
* fix the build issues resulting from the upgraded pins.

See individual commits for more detail.

I ran unit tests and quality checks locally. I plan to deploy Blockstore to production shortly after merging.

**Urgency: Moderate-low.** While not causing production issues, T&L gets alerted every week when the requirements job fails. Also, Blockstore goes months without being deployed, which isn't great (what if the pipeline stops working and we don't notice?). By getting the weekly upgrade cadence going again, we can keep version pins fresh and ensure we're deploying regularly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ html_coverage: ## Generate HTML coverage report
 
 quality: ## Run quality checks
 	${VENV_BIN}/pycodestyle --config=pycodestyle blockstore tagstore *.py
-	${VENV_BIN}/pylint --rcfile=pylintrc blockstore tagstore *.py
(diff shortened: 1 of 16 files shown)


eLRuLL:
@kdmccormick ty for this, it looks great to me, I even checked the dependency updates if they introduce backwards compatible problems, but found nothing to worry about, so I think we are good to go here :+1:

kdmccormick:
Thanks @eLRuLL !

PR #116, opened 2021-08-06
test: Adding test to check jwt encode and decode methods.

awais786:
Jwt new version `v2.0.0` has some [breaking changes](https://pyjwt.readthedocs.io/en/stable/changelog.html#dropped-deprecated-errors), this test will help to identify any issues before deployment.

Just checking encode and decodes works with params. We already saw few prod failures, this might help us little bit.

Verified this test against `v2.1.0` in this [PR](https://github.com/edx/blockstore/pull/117) and it is failing there.

### DIFF
--- a/blockstore/apps/core/tests/test_jwt.py
+++ b/blockstore/apps/core/tests/test_jwt.py
@@ -1,0 +1,62 @@
+"""
+Tests for the JWT sanity check.
+Checking edx-drf-extensinos method.
+Checking jwt.encode and jwt.decode methods.
+"""
+from time import time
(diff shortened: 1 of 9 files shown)


awais786:
@kdmccormick if you think it's valuable we can merge this.

kdmccormick:
Thanks @awais786! I am definitely open to reviewing and merging these tests.

robrap:
@awais786: It’s helpful to add a comment when a PR is closed about why it was closed. Often, it will be closed in favor of a different PR, and you can provide a link to the newer PR. Thanks.

PR #118, opened 2021-08-06
Add MIME type based file extensions to snapshot data files [SE-4663] 

### LINKED ISSUE #63: Blockstore can't serve SVG images correctly over S3
When one uploads an SVG image to a Blockstore server that's storing data on Amazon S3, Blockstore will serve the image from a URL like

```
https://our-blockstore-content.s3.amazonaws.com/ca9635e6-a43e-40df-be67-1764d4a6b311/snapshot_data/6bc0efe8f1d9722615ca1d7042b85d5af115ca6f?AWSAccessKeyId=...&Expires=1575574652
```

S3 in turn serves that response with `Content-Type: application/octet-stream`

Unfortunately, this means that one cannot use `` because browsers will not render SVG images that aren't served with the correct `Content-Type` header.

Possible solutions include:
* Change the data file name so it includes the file extension
* Have Blockstore know/detect the file type, and then [include the `response-content-type` parameter in the S3 signed URL request](https://stackoverflow.com/a/36306039), so S3 will serve it with the correct `Content-Type` header.

jvdm:
## Description

Change Draft and Version store metadata to track static file's MIME type based on its "file extension" as defined by [`mimetypes.guess_type()`](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type). Additionally, sets the extension based on the guessed mime type when saving files in the Version Store, in an attempt to allow correct MIME type headers when serving the files.

Finally, the change is backward compatible. If a file's MIME type cannot be guessed from its extension or the file doesn't have an extension at all, it is stored using the old naming schema (that is, only the file's hash). This also allows files stored with the old metadata that doesn't carry the `mime_type` field to be handled correctly.

See additional discussion in the related issue #63.

The PR approach fixes #63, but it doesn't attempt to conciliate old files with new files with the same hash content. That means, from the Blockstore perspective these are two different objects, although, in practice, they hold the same content.

## Test Instructions

1. Deploy the blockstore to a devstack as described in README.
2. Go to http://localhost:18250/api/v1
3. Use the APIs to (a.) Create a bundle, (b.) Create a draft, (c.) Then commit to the version. *Make sure to add static files*, for example, a *.jpg python.
4. Call `make blockstore-shell`.
5. Then check the following:
    - Go to `/blockstore/app/blockstore/media/<version>/snapshot_data` and verify files were added with their extension.
    - Read `/blockstore/app/blockstore/media/<version>/snapshots/<snapshot_id>.json` and verify the `content_type` flag is present.
    - Go to the files URL and verify the devstack serves the file with the right MIME type header.

## TODOs

- [ ] Squash before merging

### DIFF
--- a/blockstore/apps/bundles/store.py
+++ b/blockstore/apps/bundles/store.py
@@ -14,8 +14,9 @@ from datetime import datetime, timezone
 from pathlib import Path
 from uuid import UUID
 import codecs
-import logging
 import json
+import logging
+import mimetypes
 import pytz
 
 from django.core.files.base import ContentFile, File
@@ -42,6 +43,14 @@ class FileInfo:
     public = attr.ib(type=bool)
     size = attr.ib(type=int)
     hash_digest = attr.ib(type=bytes)
+    mime_type = attr.ib(type=str, default=None)
+
+    @property
+    def data_filename(self):
+        """The data filename with an extension, if present."""
+        hash_hex = self.hash_digest.hex()
+        extension = mimetypes.guess_extension(self.mime_type) if self.mime_type else ''
+        return f"{hash_hex}{extension}"
 
     @staticmethod
     def generate_hash(django_file):
@@ -50,6 +59,12 @@ class FileInfo:
         for chunk in django_file.chunks():
             hash_obj.update(chunk)
         return hash_obj
+
+    @staticmethod
+    def guess_mime_type(path):
+        """Given a path, return its IANA MIME type."""
+        # Return only the 'type/subtype', usable for a MIME content-type header.
+        return mimetypes.guess_type(path)[0]
 
     @classmethod
     def from_json_dict(cls, json_dict):
@@ -60,13 +75,24 @@ class FileInfo:
                 # Used by Drafts to indicate this file should be deleted.
                 file_info_dict[pathname] = None
             else:
-                file_info_dict[pathname] = cls(
+                kwds = dict(
                     path=pathname,
                     public=file_info[0],
                     size=file_info[1],
                     hash_digest=bytes_from_hex_str(file_info[2])
                 )
+                if len(file_info) > 3:
+                    kwds['mime_type'] = file_info[3]
+                file_info_dict[pathname] = cls(**kwds)
         return file_info_dict
+
+    def to_json_object(self):
+        """Return a python object suitable for json serialization."""
+        fields = [self.public, self.size, self.hash_digest.hex()]
+        # If no MIME type don't write nulls.
+        if self.mime_type:
+            fields.append(self.mime_type)
+        return fields
 
 
 @attr.s(frozen=True)
@@ -236,14 +262,25 @@ class SnapshotRepo:
     Metadata JSON file describing the contents of a Snapshot and basic metadata:
         {bundle_uuid}/snapshots/{snapshot_digest.hex()}.json
 
-    Data for each file:
+    Data for each file can be stored with or without an extension, based on the
+    existance of a ``mime_type`` attribute:
+
         {bundle_uuid}/snapshot_data/{file_hash.hexdigest()}
+
+    Or (e.g. image/jpeg):
+
+        {bundle_uuid}/snapshot_data/{file_hash.hexdigest()}.jpg
+
+    The extension is infered from the original static file extension, see
+    ``mimetypes.guess_type()``.
 
     Notes:
     * To make S3 partitioning work correctly, we need to lead with the UUID or
       some other randomly distributed string.
     * A data file is stored with a hash as a name since it could be referenced
       by different paths in different versions.
+    * The data file extension allows underlying Persistance/Datastore layer to
+      serve the file with the correct MIME content-type header.
     """
     def __init__(self, storage=None):
         self.storage = storage or default_asset_storage
@@ -270,6 +307,8 @@ class SnapshotRepo:
                 f"Snapshot {snapshot_digest.hex()} for Bundle {bundle_uuid} not found"
             ) from err
 
+        # TODO Can we use a more robust JSON schema handling here, and take
+        #      advantage of _version?
         return Snapshot(
             bundle_uuid=bundle_uuid,
             files=FileInfo.from_json_dict(snapshot_json['files']),
@@ -296,12 +335,12 @@ class SnapshotRepo:
     def url(self, snapshot, path):
         """Return a user-accessible URL to download a path from this Snapshot."""
         file_info = snapshot.files[path]
-        storage_path = self._file_data_path(snapshot.bundle_uuid, file_info.hash_digest.hex())
+        storage_path = self._file_data_path(snapshot.bundle_uuid, file_info)
         return self.storage.url(storage_path)
 
     def open(self, snapshot, path):
         file_info = snapshot.files[path]
-        storage_path = self._file_data_path(snapshot.bundle_uuid, file_info.hash_digest.hex())
+        storage_path = self._file_data_path(snapshot.bundle_uuid, file_info)
         return self.storage.open(storage_path, 'rb')
 
     @classmethod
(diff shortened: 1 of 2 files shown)


edx-webhook:
Thanks for the pull request, @jvdm! I've created [OSPR-5953](https://openedx.atlassian.net/browse/OSPR-5953) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket:
- supporting documentation

This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

natabene:
@jvdm Thank you for your contribution. Please let me know once this is ready for a review.

jvdm:
Thank you, @natabene. I believe this is ready for review. Although, some discussion may happen at #63 regarding the approach taken here.

bradenmacdonald:
@ormsbee Would you like to review this? It is mostly touching on the code that you originally wrote I believe.

I am happy to review in addition or instead.

blockstore/apps/bundles/tests/test_store.py
```diff
@@ -88,12 +112,14 @@ def test_basic_creation(self, snapshot_created_mock):
...
                     size=25,
                     hash_digest=create_hash(HTML_CONTENT_BYTES).digest(),
+                    mime_type='text/html',
                 ),
                 'test.txt': FileInfo(
                     path='test.txt',
                     public=False,
                     size=17,
                     hash_digest=create_hash(TEXT_CONTENT_BYTES).digest(),
+                    mime_type='text/plain',
```
arbrandes, line 122: @jvdm, do you mind adding a file (or two) here that fall under the situation [Braden originally described](https://github.com/edx/blockstore/issues/63#issuecomment-893870623) where "two files are uploaded with different extensions but the same content"?  In particular, when such files resolve to the exact same MIME type.

If for nothing else other than to document that this was foreseen, and to protect this particular case from future bugs.  I couldn't detect a related problem just by reading the code.

arbrandes (changes requested):
@jvdm, I generally approve of the PR.  I tested it in a devstack with real data, and it solves the problem as originally reported.  Congrats, and thanks!

I just asked for one change regarding tests.

arbrandes:
Also, can you please add [SE-4663] to the title, to aid our internal tracking?

blockstore/apps/bundles/tests/test_store.py
jvdm: @arbrandes 

> do you mind adding a file (or two) here that fall under the situation Braden originally described where "two files are uploaded with different extensions but the same content"?

Can do. Although, I want to clarify what will happen.

> In particular, when such files resolve to the exact same MIME type.

Two scenarios, (a.) If the different extensions are in fact the same MIME, the files will dedup in the storage layer, just like today. That is correct because the extensions are the source of truth for the MIME-type header in the patch proposed, so one potential example is `cat.jpeg` and `samecat.jpg` will be stored as `<some-hash>.jpg`. Now, (b.) if different extensions are guessed to be two different MIME types, for example, if you do `mv cat.jpg cat.pdf` and upload `cat.pdf` and `cat.jpeg`, blockstore will create two files with the same content.

You might argue that for (b.), we would be better using the files' content as the source of the truth for the MIME header. That is possible, but in my opinion, not as flexible as using the file extension. One example, what if the content-based MIME type guessing fails for some (I know, very particular) case? Customers will feel confused, as in "Why my PDF is opening as an image?". And without a clear path to move forward. So, unless we add some additional API parameters to bypass the guessing, I prefer the file extension. Let me know if you want to explore the API parameter to select the type forcibly.

bradenmacdonald: For what it's worth, I think what you're describing in cases (a) and (b) sound good and I was thinking along the same lines too.

bradenmacdonald:
@natabene Can I review this as core committer?

bradenmacdonald:
@arbrandes Did you also test backwards compatibility ?

blockstore/apps/bundles/store.py
```diff
@@ -42,6 +43,14 @@ class FileInfo:
     public = attr.ib(type=bool)
     size = attr.ib(type=int)
     hash_digest = attr.ib(type=bytes)
+    mime_type = attr.ib(type=str, default=None)
+
+    @property
+    def data_filename(self):
+        """The data filename with an extension, if present."""
+        hash_hex = self.hash_digest.hex()
+        extension = mimetypes.guess_extension(self.mime_type) if self.mime_type else ''
```
bradenmacdonald, line 52: I think I see the potential for a rare but possible problem here:

Scenario:
1. User uploads a video file; it's `mime_type` is stored as `video/mp4` and its data file is stored at `[hash].mp4`
2. We update python (or the underlying OS used in the docker image, etc.), and `mimetypes` now returns `.m4v` as the file extension for `video/mp4`
3. Blockstore hasn't stored the file extension, only the mimetype, so now looks for `[hash].mp4` which doesn't exist.


Obviously this isn't super likely, but it has happened before with mimetype projects, e.g. [this change of extension for `audio/mp4` from `.mp4a` to `.m4a`](https://github.com/jshttp/mime-db/commit/ae413e174e9b8d3026b39a43a9a044b72be41ab8) or [this very recent change of the extension for `application/ecmascript` from `.ecma` to `.es`](https://github.com/jshttp/mime-db/commit/03d57a35d71f5897f3a2a9df488c6ef76f34ccc6)

Another related issue is that `mimetypes` can return different results on different systems since it uses data files from the underlying OS.

All that said, it's probably not really a problem in practice and we could always add blockstore-specific overrides for certain mimetypes here if it ever was a problem. Thoughts?

bradenmacdonald:
Code looks good to me, just the one question above. If @arbrandes has carefully tested backwards compatibility I'll approve this; otherwise I will as soon as he or I has done so.

arbrandes:
@bradenmacdonald, I have just deployed this branch to a pre-existing sandbox that has Blockstore using MinIO and a large number of previously-imported blocks, and could detect no ill effects.  Previously uploaded SVGs still don't work, of course, but other assets such as images, and otherwise the bundle itself, are all fine.

bradenmacdonald (approved):

bradenmacdonald:
Please squash to a single commit and we can merge next week if there are no further comments or concerns.

blockstore/apps/bundles/store.py
jvdm: @bradenmacdonald:

> I think I see the potential for a rare but possible problem here:
>
> [...]
>
> Blockstore hasn't stored the file extension, only the mimetype, so now looks for [hash].mv4 which doesn't exist.

Oh, that's a nice catch! Loved it. :heart: 

> Thoughts?

This dependency between blockstore behavior and its environment will be painful to maintain. A deployment or upgrade could easily break multiple static file URLs. For example, `XBlockRuntime` relies on [blockstore to translate static URLs](https://github.com/edx/edx-platform/blob/49aaa8621d195a452000942aae8286966a91b96a/openedx/core/djangoapps/xblock/runtime/blockstore_runtime.py#L168) which in turn will use the [snapshot URL generation](https://github.com/edx/blockstore/blob/ece89fa79cd4e4a320f2d8c977032af0a52e2680/blockstore/apps/api/v1/serializers/bundles.py#L152) to retrieve it, and if the extension guessed changes, let's say, because of deployment, so does the static file URL, yielding 400's.

It should be relatively simple to extend the patch to add the "guessed file extension" to the snapshot metadata. But, at this point, I think the patch will be more resilient if we go all the way and save the whole "snapshot filename" instead. Changes to the naming schema won't affect previous Snapshots if we rely on a static filename definition in the snapshot metadata.

Let's get a consensus on this before I submit a new revision, @bradenmacdonald.

ormsbee:
> @ormsbee Would you like to review this? It is mostly touching on the code that you originally wrote I believe.

Sorry for the delay in reply. I'm on PTO for the next week. I trust you folks. Thank you.

jvdm:
> Please squash to a single commit and we can merge next week if there are no further comments or concerns.

Rebased onto `origin/master` and squashed.

blockstore/apps/bundles/store.py
bradenmacdonald: @jvdm Whoops sorry for the delayed reply here. 

Do you want to merge this now or address this first? I'm fine either way. Although it does seem like a risk of a bug in the future, the odds strike me as very low, and the mitigation would be quite easy (`if mimetype='problematic/type': file_ext = ".forced"; else: file_ext = guess_mime_type()`)

>  I think the patch will be more resilient if we go all the way and save the whole "snapshot filename" instead

Sounds good to me

jvdm: @bradenmacdonald:

> Do you want to merge this now or address this first? I'm fine either way.

Let's merge. We can unblock customers now and follow up on this in a separate PR.

bradenmacdonald: OK I'll merge tomorrow so that it deploys during edX business hours.

edx-webhook:
@jvdm 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/blockstore&entry.867055334=https://github.com/edx/blockstore/pull/118&entry.1484655318=https://github.com/jvdm&entry.752974735=2021-08-06+09:14&entry.1917517419=2021-08-19+15:24&entry.2133058324=Yes) so we can improve your experience in the future.

PR #119, opened 2021-08-10
feat!: upgrade pyjwt to 2.1.0

iamsobanjaved:
**Breaking changes.**

In blockstore with latest version of pyjwt=2.1.0 this error appears.
`It is required that you pass in a value for the "algorithms" argument when calling decode().`

Its due to this change.
https://pyjwt.readthedocs.io/en/stable/changelog.html#dropped-deprecated-verify-expiration-param-in-jwt-decode


awais786:
@kdmccormick Please review and test this PR.

Makefile
```diff
@@ -131,8 +131,19 @@ docker_push: docker_tag docker_auth ## push to docker hub
...
 
+COMMON_CONSTRAINTS_TXT=requirements/common_constraints.txt
+.PHONY: $(COMMON_CONSTRAINTS_TXT)
+$(COMMON_CONSTRAINTS_TXT):
+	wget -O "$(@)" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt || touch "$(@)"
+
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
-upgrade: ## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
+upgrade: $(COMMON_CONSTRAINTS_TXT)	## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
+	sed 's/pyjwt\[crypto\]<2.0.0//g' requirements/common_constraints.txt > requirements/common_constraints.tmp
```
iamsobanjaved, line 141: We are creating a local copy of the common constraints file and override/remove the global constraint which we don't need. This approach is used temporarily and will be improved in [BOM-2721](https://openedx.atlassian.net/browse/BOM-2721)

awais786:
@kdmccormick Please review this.

kdmccormick:
@doctoryes Heads up on this review if you have time today. If not, IMO this can sit until next week.

Makefile
doctoryes, line 141: Should a first step and last step be added here to copy the *existing* `requirements/common_constraints.txt` file into a temp file and then restore it after all pip-compiles have run? Otherwise, a `make upgrade` will always modify the common constraints file according to a `git status`?

doctoryes (commented):
The proposed technique seems fine temporarily. Just one question....

Makefile
iamsobanjaved: The `requirements/common_constraints.txt` file will only get updated if there is some change in the actual constraint file here (https://github.com/edx/edx-lint/blob/master/edx_lint/files/common_constraints.txt)

doctoryes (approved):

PR #135, opened 2021-10-25
Make blockstore an installable package

symbolist:
## Description

This PR pull out the following changes from https://github.com/edx/blockstore/pull/97:
* Move apps/api to apps/rest_api.
* Copy the latest code in `edx-platform/openedx/core/lib/blockstore_api` to `blockstore/apps/api`. This has been done to make the review of https://github.com/edx/blockstore/pull/97 easier since we will get a useful diff of changes made to each of the API methods.
* Makes the apps into an installable package.

It also disables the migration code which changes the MySQL character set to `utfmb4` to prevent accidentally running it against an edxapp database.

## Ticket

* [FAL-170](https://tasks.opencraft.com/browse/FAL-170)

## Test Instructions

* Checkout branch.
* In the Open edX devstack run:
```
make dev.shell.mysql57
mysql -u root
DROP DATABASE blockstore_db;
```

* Start and provision the blockstore docker container.
* Run tests with `make test`.
* Start the blockstore django server and check the API is working correctly using the Libraries UI or LabXchange devstack.

Add `'blockstore.apps.bundles.apps.BundlesConfig',` to `lms/envs/common.py::INSTALLED_APPS`.

Install the package in LMS:
```
make lms-shell
pip install -e /edx/src/blockstore
./manage.py lms migrate
```

Check the API methods can be called:
```
./manage.py lms shell
from blockstore.api import *
create_collection("Collection") # This should fail with requests.exceptions.ConnectionError
```

## TODOs

None

PR #140, opened 2021-11-16
Make blockstore an installable package [SE-4987]

jvdm:
## Description

This is a follow-up PR built on top of @symbolist's #135. It's still part of #97, and is a step towards its completion.

The previous description holds, and is copied verbatim here for reference:

- Move apps/api to apps/rest_api.
- Copy the latest code in edx-platform/openedx/core/lib/blockstore_api to blockstore/apps/api. This has been done to make the review of Blockstore as App [SE-3321] #97 easier since we will get a useful diff of changes made to each of the API methods.
- Makes the apps into an installable package.
- It also disables the migration code which changes the MySQL character set to utfmb4 to prevent accidentally running it against an edxapp database.

Co-authored-by: Usman Khalid <2200617@gmail.com>

## Test Instructions

In your devstack:

- Go to your `devstack`, enable [frontend-app-library-authoring](https://github.com/edx/frontend-app-library-authoring), put everything up.
- Drop any potential blockstore tables, see:

  ```
  make dev.shell.mysql57
  mysql -u root
  DROP DATABASE blockstore_db;
  ```
  
- Provision blockstore by following its README steps.  This should create the tables from `master`.
- Run tests with `make test`.
- Go to `localhost:3001`, create a library.
- Create any block inside the library.
- Got to `blockstore` and stop blockstore development service.
- Checkout the development branch `jvdm/packagify-blockstore`.
- Provision `blockstore` one more time.
- Open `localhost:3001` and verify the library and block opens successfully.
- Add `blockstore.apps.bundles.apps.BundlesConfig`, to `lms/envs/common.py::INSTALLED_APPS`.
- Install the package in LMS:

  ```
  make dev.shell.lms
  pip install -e /edx/src/blockstore
  python manage.py lms migrate
  ```

- Open the django shell `python manage.py lms shell` and check the API methods
  can be called:

  ```
  from blockstore.api import *
  create_collection("Collection") # This should fail with requests.exceptions.ConnectionError
  ```


edx-webhook:
Thanks for the pull request, @jvdm! I've created [OSPR-6227](https://openedx.atlassian.net/browse/OSPR-6227) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

kdmccormick:
Thanks for the PR @jvdm @jristau1984 - I'll leave a review this week.

setup.py
```diff
@@ -0,0 +1,92 @@
...
+    include_package_data=True,
+    install_requires=load_requirements('requirements/base.in'),
+    python_requires=">=3.8",
+    license="AGPL 3.0",
+    zip_safe=False,
+    keywords='Python edx',
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Framework :: Django',
+        'Framework :: Django :: 2.2',
```
ormsbee, line 85: Shouldn't this be Django 3.2?

blockstore/api.py
```diff
@@ -0,0 +1,6 @@
+"""
+The Blockstore Python API.
+
+Imported here to allow consumers to import the public api from `blockstore.api`.
+"""
+from blockstore.apps.api import *  # pylint: disable=wildcard-import
```
ormsbee, line 6: Is this necessary to do? These API calls return models that echo models that are in blockstore itself. Also, it's inconsistent with more recent practice in [OEP-49](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0049-django-app-patterns.html#data-py) to create a separate set of `attrs` data structures to pass back in these situations.

I realize this is an intermediate state, and I imagine a lot of the models brought over from edx-platform's client-side code. My main concern is that we'd be introducing an api module with some implied level of stability, and then possibly changing all of it.

Is the plan for what happens to the API redundancies written down somewhere? I'm fine with this as an intermediate state, but I'd like to understand where we're going with collapsing these down?

ormsbee (changes requested):
Minor bit about Django versions, and some questions about what the final state is going to be.

Another question: What was the root cause of #135 getting reverted (#138)?

Thank you.

blockstore/api.py
symbolist: @ormsbee 

I agree that eventually we should only have one set of `attrs`. However, the problem is that there are differences between the two sets of attrs. The easiest way to allow fast rollbacks and to keep the changes easy to understand was to keep them all in the `methods.api` file. If we try to collapse them now, it would require making changes to callers in various places in edx-platform and make rollbacks harder. This way ensures that from the point of view of callers, the API hasn't changed. And once blockstore is an in-process app, we can incrementally update the APIs and their callers.

> My main concern is that we'd be introducing an api module with some implied level of stability, and then possibly changing all of it.

Agree with this. @jvdm We should remove this file for now and only add methods later that we want to mark stable.

> Is the plan for what happens to the API redundancies written down somewhere? I'm fine with this as an intermediate state, but I'd like to understand where we're going with collapsing these down?

Nope, not yet. I was thinking we would look at the usages in edx-platform and then see how the APIs should be updated for better performance and ergonomics – the system boundary has changed and there may be opportunities for these?

symbolist:
@jvdm Can you please keep the original commits? They are different changes and have separate commit messages so keeping them separate is important.

symbolist:
@ormsbee

> Another question: What was the root cause of #135 getting reverted (#138)?

Hmm, I left a comment somewhere but am unable to find it now. Basically I had [updated](https://github.com/edx/blockstore/pull/135/files#diff-afc7bc0243927d117e5adeb82d8e0e62bab869608c14ca9dfb19bf0d9d85bddeR8) the label of the bundles app to namespace it but forgot it would cause the models to assume a different table name (and none of the reviewers caught that because everyone was testing with a fresh db). 🙂

setup.py
jvdm: +1

blockstore/api.py
jvdm: > @jvdm We should remove this file for now and only add methods later that we want to mark stable.

+1

jvdm:
@ormsbee Leaving a ping here, we have updated the PR with the requested changes you left. Let us know if you would like to follow up on anything else. Thank you.

blockstore/api.py
ormsbee: >  I was thinking we would look at the usages in edx-platform and then see how the APIs should be updated for better performance and ergonomics – the system boundary has changed and there may be opportunities for these?

Sounds good.

ormsbee (approved):
I'm approving the changes with two minor caveats:

1. Please change the commit message to include more of the context from the PR message.
2. I don't think I have the access to deploy this any longer, so I'm tagging @kenclary and @connorhaugh on that front.

jvdm:
@ormsbee:

> Please change the commit message to include more of the context from the PR message.

Done.

> I don't think I have the access to deploy this any longer, so I'm tagging @kenclary and @connorhaugh on that front.

Cool, thanks.

kdmccormick:
@jvdm , once this merges, what remains to be done in https://github.com/edx/blockstore/pull/97 ?

edx-webhook:
@jvdm 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/blockstore&entry.867055334=https://github.com/edx/blockstore/pull/140&entry.1484655318=https://github.com/jvdm&entry.752974735=2021-11-16+21:33&entry.1917517419=2021-11-29+19:59&entry.2133058324=Yes) so we can improve your experience in the future.

connorhaugh:
This seems to have deployed with no noticeable errors.

symbolist:
@kdmccormick 

> once this merges, what remains to be done in #97 ?

Steps 2-4 mentioned in https://tasks.opencraft.com/browse/EDXCL-4 🙂

kdmccormick:
Thanks @symbolist , I think that ticket requires a tasks.opencraft.com account -- would I be able to get an account?

jristau1984:
Steps 3 and 4 from the ticket above copied here for reference: 

3 - There are a few tests which are integration tests but need to become unit tests (since blockstore is going to be an installed app). These need to be verified and any fixes made.

4 - There is a bug related to how the media urls for files in bundles are passed around in the devstack that needs to be resolved.

kdmccormick:
Thanks @jristau1984 !

It seems like EDXCL-4's Step 2 corresponds to [Step 2 from the TNL migration plan](https://openedx.atlassian.net/browse/TNL-8746), which is great.

Heads up that the TNL plan asks for Step 2 to be done behind a feature flag, and that it requires [Step 1 from the TNL plan (giving edxapp access to the blockstore DB)](https://openedx.atlassian.net/browse/TNL-8745) to be done first.

PR #149, opened 2022-01-13
 [BD-14] Updates Blockstore Python API [TNL-8746] [FAL-2702]

pomegranited:
## Description

Continues the work started by https://github.com/openedx/blockstore/pull/140, and makes the remaining changes necessary for the Blockstore as app ADR: [#71](https://github.com/openedx/blockstore/pull/71).

Replaces https://github.com/openedx/blockstore/pull/97.

- [x] Update the Python API to use the models directly.
- [x] Enhance unit tests to improve coverage (from 85% to 98%)

See https://github.com/openedx/blockstore/pull/97 for discussions.

## Test Instructions

See https://github.com/openedx/edx-platform/pull/29779.

## TODOs

- [ ] Update README to reflect these changes, specifically the [blockstore overview](https://github.com/openedx/blockstore#blockstore) and [Running integration tests](https://github.com/openedx/blockstore#running-integration-tests).
- [ ] Squash before merging
- [ ] Tag the new release version.

### Next steps

Once these PRs merge, we will need to:

Clean up the edx-platform codebase:

- [ ] Remove the waffle flag added by https://github.com/openedx/edx-platform/pull/29779.
- [ ] Replace the `openedx.core.lib.blockstore_api` with `blockstore.app.api`.
- [ ] Copy over the data from the blockstore database to the edxapp database.
- [ ] Remove the database router added above.

Clean up the blockstore codebase:

- [ ] Remove [`blockstore.apps.core`](https://github.com/openedx/blockstore/tree/master/blockstore/apps/core).
- [ ] Remove the [`blockstore.app.rest_api`](https://github.com/openedx/blockstore/tree/master/blockstore/apps/rest_api)
- [ ] Remove remaining [`tagstore`](https://github.com/openedx/blockstore/tree/master/tagstore) code.

### DIFF
--- a/blockstore/apps/api/methods.py
+++ b/blockstore/apps/api/methods.py
@@ -76,7 +76,7 @@ def get_bundles(uuids=None, text_search=None):
     """
     Get the details of all bundles.
     """
-    bundles_queryset = models.Bundle.objects.all()
+    bundles_queryset = _bundle_queryset()
     if uuids:
         bundles_queryset = bundles_queryset.filter(uuid__in=uuids)
     if text_search:
@@ -154,7 +154,7 @@ def get_or_create_bundle_draft(bundle_uuid, draft_name):
     Retrieve metadata about the specified draft, creating a new one if it does not exist yet.
     """
     try:
-        draft_model = models.Draft.objects.get(bundle__uuid=bundle_uuid, name=draft_name)
+        draft_model = _draft_queryset().get(bundle__uuid=bundle_uuid, name=draft_name)
     except models.Draft.DoesNotExist:
         bundle_model = _get_bundle_model(bundle_uuid)
         draft_model = models.Draft(
@@ -201,6 +201,8 @@ def get_bundle_version(bundle_uuid, version_number=None):
     Get the details of the specified bundle version
     """
     bundle_version_model = _get_bundle_version_model(bundle_uuid, version_number)
+    if bundle_version_model is None:
+        return None
     return _bundle_version_data_from_model(bundle_version_model)
 
 
@@ -209,8 +211,12 @@ def get_bundle_version_files(bundle_uuid, version_number):
     Get a list of the files in the specified bundle version
     """
     if version_number == 0:
+        # There are no files in the initial version of a bundle
         return []
-    return get_bundle_version(bundle_uuid, version_number).files.values()
+    bundle_version = get_bundle_version(bundle_uuid, version_number)
+    if bundle_version:
+        return bundle_version.files.values()
+    return []
 
 
 def get_bundle_version_links(bundle_uuid, version_number):
@@ -218,26 +224,33 @@ def get_bundle_version_links(bundle_uuid, version_number):
     Get a dictionary of the links in the specified bundle version
     """
     if version_number == 0:
+        # There are no links in the initial version of a bundle
         return {}
-    return get_bundle_version(bundle_uuid, version_number).links
+    bundle_version = get_bundle_version(bundle_uuid, version_number)
+    if bundle_version:
+        return bundle_version.links
+    return {}
 
 
 def get_bundle_files_dict(bundle_uuid, use_draft=None):
     """
-    Get a dict of all the files in the specified bundle.
+    Get a dict of all the files in the specified bundle or draft.
 
     Returns a dict where the keys are the paths (strings) and the values are
     BundleFileData or DraftFileData tuples.
     """
     if use_draft:
         try:
-            draft_model = models.Draft.objects.get(bundle__uuid=bundle_uuid, name=use_draft)
+            draft_model = _draft_queryset().get(bundle__uuid=bundle_uuid, name=use_draft)
         except models.Draft.DoesNotExist:
             pass
         else:
             return _draft_data_from_model(draft_model).files
 
-    return get_bundle_version(bundle_uuid).files
+    bundle_version = get_bundle_version(bundle_uuid)
+    if bundle_version:
+        return bundle_version.files
+    return {}
 
 
 def get_bundle_files(bundle_uuid, use_draft=None):
@@ -256,13 +269,16 @@ def get_bundle_links(bundle_uuid, use_draft=None):
     """
     if use_draft:
         try:
-            draft_model = models.Draft.objects.get(bundle__uuid=bundle_uuid, name=use_draft)
+            draft_model = _draft_queryset().get(bundle__uuid=bundle_uuid, name=use_draft)
         except models.Draft.DoesNotExist:
             pass
         else:
             return _draft_data_from_model(draft_model).links
 
-    return get_bundle_version(bundle_uuid).links
+    bundle_version = get_bundle_version(bundle_uuid)
+    if bundle_version:
+        return get_bundle_version(bundle_uuid).links
+    return {}
 
 
 def get_bundle_file_metadata(bundle_uuid, path, use_draft=None):
@@ -288,7 +304,7 @@ def get_bundle_file_data(bundle_uuid, path, use_draft=None):
 
     if use_draft:
         try:
-            draft_model = models.Draft.objects.get(bundle__uuid=bundle_uuid, name=use_draft)
+            draft_model = _draft_queryset().get(bundle__uuid=bundle_uuid, name=use_draft)
         except models.Draft.DoesNotExist:
             pass
         else:
@@ -297,7 +313,7 @@ def get_bundle_file_data(bundle_uuid, path, use_draft=None):
             with draft_repo.open(staged_draft, path) as file:
                 return file.read()
 
-    bundle_version_model = _get_bundle_version_model(bundle_uuid)
+    bundle_version_model = _get_bundle_version_model(bundle_uuid, 0)
 
     snapshot_repo = SnapshotRepo()
     snapshot = bundle_version_model.snapshot()
@@ -391,6 +407,15 @@ def _collection_data_from_model(collection_model):
     return CollectionData(uuid=collection_model.uuid, title=collection_model.title)
 
 
+def _bundle_queryset():
+    """
+    Returns the bundle model queryset.
+
+    Prefetch the data needed to create BundleData objects.
+    """
+    return models.Bundle.objects.prefetch_related('drafts', 'versions')
+
+
 def _get_bundle_model(bundle_uuid):
     """
     Get Bundle model from UUID.
@@ -398,7 +423,7 @@ def _get_bundle_model(bundle_uuid):
     Raises BundleNotFound if bundle with UUID does not exist.
     """
     try:
-        bundle_model = models.Bundle.objects.get(uuid=bundle_uuid)
+        bundle_model = _bundle_queryset().get(uuid=bundle_uuid)
     except models.Bundle.DoesNotExist as exc:
         raise BundleNotFound("Bundle {} does not exist.".format(bundle_uuid)) from exc
     return bundle_model
@@ -408,15 +433,26 @@ def _bundle_data_from_model(bundle_model):
     """
     Create and return BundleData from bundle model.
     """
-    latest_bundle_version_model = bundle_model.get_bundle_version()
+    latest_version = bundle_model.versions.order_by('-version_num').first()
+    latest_version_num = latest_version.version_num if latest_version else 0
+
     return BundleData(
         uuid=bundle_model.uuid,
         title=bundle_model.title,
         description=bundle_model.description,
         slug=bundle_model.slug,
         drafts={draft.name: draft.uuid for draft in bundle_model.drafts.all()},
-        latest_version=latest_bundle_version_model.version_num if latest_bundle_version_model else 0,
+        latest_version=latest_version_num,
     )
+
+
+def _draft_queryset():
+    """
+    Returns the draft model queryset.
+
+    Prefetch the data needed to create DraftData objects.
+    """
+    return models.Draft.objects.select_related('bundle')
 
 
 def _get_draft_model(draft_uuid):
@@ -426,7 +462,7 @@ def _get_draft_model(draft_uuid):
     Raises DraftNotFound if draft with UUID does not exist.
     """
     try:
-        draft_model = models.Draft.objects.get(uuid=draft_uuid)
+        draft_model = _draft_queryset().get(uuid=draft_uuid)
     except models.Draft.DoesNotExist as exc:
         raise DraftNotFound("Draft {} does not exist.".format(draft_uuid)) from exc
     return draft_model
@@ -475,6 +511,15 @@ def _draft_data_from_model(draft_model):
     )
 
 
+def _bundle_version_queryset():
+    """
+    Returns the bundle version model queryset.
+
+    Prefetch the data needed to create BundleVersionData objects.
+    """
+    return models.BundleVersion.objects.select_related('bundle')
+
+
 def _get_bundle_version_model(bundle_uuid, version_number=None):
     """
     Get BundleVersion from bundle UUID and version number.
@@ -487,8 +532,8 @@ def _get_bundle_version_model(bundle_uuid, version_number=None):
     if version_number:
         filter_kwargs['version_num'] = version_number
 
-    bundle_version_model = models.BundleVersion.objects.filter(**filter_kwargs).order_by('-version_num').first()
-    if bundle_version_model is None:
+    bundle_version_model = _bundle_version_queryset().filter(**filter_kwargs).order_by('-version_num').first()
+    if version_number and bundle_version_model is None:
         raise BundleVersionNotFound("Bundle Version {},{} does not exist.".format(bundle_uuid, version_number))
     return bundle_version_model
 
--- a/blockstore/apps/api/methods.py
+++ b/blockstore/apps/api/methods.py
@@ -3,6 +3,7 @@ API Client methods for working with Blockstore bundles and drafts
 """
 
 import base64
+import re
 import crum
 
 from django.db.models import Q
@@ -214,9 +215,7 @@ def get_bundle_version_files(bundle_uuid, version_number):
         # There are no files in the initial version of a bundle
         return []
     bundle_version = get_bundle_version(bundle_uuid, version_number)
-    if bundle_version:
-        return bundle_version.files.values()
-    return []
+    return list(bundle_version.files.values() if bundle_version else [])
 
 
 def get_bundle_version_links(bundle_uuid, version_number):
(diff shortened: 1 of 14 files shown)


edx-webhook:
Thanks for the pull request, @pomegranited! I've created [OSPR-6383](https://openedx.atlassian.net/browse/OSPR-6383) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

natabene:
@pomegranited Thank you for your contribution. Please let me know once this is ready.

blockstore/apps/bundles/models.py
```diff
@@ -54,13 +54,22 @@
...
 
 MAX_CHAR_FIELD_LENGTH = 180
 
 
+"""
+Sqlite is used for testing on edx-platform, but doesn't contain the utf8mb4_general_ci collation sequence.
+
+So we detect this case here, and use a different db_collation if we're running on Sqlite.
+"""
+DB_COLLATION = 'binary' if 'sqlite' in settings.DATABASES['default']['ENGINE'] else 'utf8mb4_general_ci'
```
pomegranited, line 70: @symbolist @gabor-boros  There's a database compatibility problem with the `db_collation` setting that was chosen for some of the fields in these models.

Blockstore tests and production use MySQL, so using `db_collation='utf8mb4_general_ci'` was fine.
However, my attempts to run the unit tests with blockstore installed have failed because edx-platform uses sqlite.

* Sqlite, doesn't support `utf8mb4_general_ci`; the best it can do is [`binary`](https://www.sqlite.org/datatype3.html#collating_sequences).
* Postgres doesn't support `utf8mb4_general_ci` either, and its use contradicts the [Postgres compatibility as documented](https://github.com/openedx/blockstore/pull/149/files#diff-2d364908a68c64f2a417d1036d2c3e5f95d8c4e1f585757818ee31b09ce79564R20-R240) for these models. Postgres's closest built-in equivalent collation sequence is the [standard `ucs_basic`](https://www.postgresql.org/docs/10/collation.html#id-1.6.10.4.5.5).

While it's relatively easy to configure the charset and collation sequence options at the `settings.DATABASES` level, we don't want to do this for all fields, and there [isn't a nice way to do this per-field](https://stackoverflow.com/questions/59503476/setting-collation-and-charset-per-model-table-in-django-and-mysql), except by modifying the model code directly.

So a full solution would require:
* detecting [which database is in use for the given class](https://stackoverflow.com/a/30459399/4302112), and 
* returning an appropriate collation setting for that database engine.

All of which makes the model.py and the generated migrations pretty complicated, and difficult to reliably test and maintain.

@ormsbee has raised the very timely issue of whether we can formally [drop support for Postgres](https://discuss.openedx.org/t/postgresql-and-open-edx/6441), but that's a longer discussion.

Any suggestions for how to handle this better right now?

ormsbee: Does it make sense to create a custom field that subclasses `CharField` and has the logic to apply database-specific collations during migrations? I think it's primarily two use cases:

* Human-readable text that should be collated to be case-insensitive and as aware of accents, etc. as possible.
* Identifiers, which should be case sensitive.

blockstore/apps/api/methods.py
```diff
@@ -420,18 +371,154 @@ def encode_str_for_draft(input_str):
...
+
+    return BundleVersionData(
+        bundle_uuid=bundle_version_model.bundle.uuid,
+        version=bundle_version_model.version_num,
+        change_description=bundle_version_model.change_description,
+        created_at=snapshot.created_at,
+        files={
+            path: BundleFileData(
+                path=path,
+                url=_build_absolute_uri(snapshot_repo.url(snapshot, path)),
```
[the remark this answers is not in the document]
symbolist: Do be honest my memory of the bug is quite fuzzy. Do this change allow any images added to the HtmlBlock to be  displayed in the browser correctly in the read view?

blockstore/apps/bundles/models.py
bradenmacdonald: I _think_ the tests don't depend on unicode-aware cast insensitive sorting, so `binary` should work fine for test purposes? The main thing they test is that all unicode characters including emojis can be used without issue, and `binary` should be fine for that.

And yeah +1 to what Dave is suggesting.

pomegranited: @bradenmacdonald 

> I think the tests don't depend on unicode-aware cast insensitive sorting, 

Yep, this hack works for testing, but it's not pretty, and doesn't handle the Postgres use case.

> Does it make sense to create a custom field that subclasses CharField and has the logic to apply database-specific collations during migrations?

Interesting suggestion, @ormsbee.. a custom field class seems doable, but it will still need to be passed information about the database selected for the migration so that it can exhibit the best field options for that engine. I'd have to dig into Django's field and migration logic to see if that can be done with the existing auto-generated migrations or not, but it's worth a try!

Since this is likely something that would be broadly useful outside of Blockstore, is there an `openedx` repo that would be an appropriate place for this change to live?

I'm going to have to leave this hack in place right now though, and create a follow-up task for this, since it's out of scope for this change.

ormsbee: > Since this is likely something that would be broadly useful outside of Blockstore, is there an openedx repo that would be an appropriate place for this change to live?

[edx-django-utils](https://github.com/openedx/edx-django-utils) is probably what you want. It's installed on pretty much every service.

blockstore/apps/bundles/storage.py
```diff
@@ -90,12 +90,18 @@ def __init__(self):
...
+        # All other S3 settings will be pulled in automatically from Django settings
+        # (such as AWS_QUERYSTRING_EXPIRE and AWS_STORAGE_BUCKET_NAME).
+        s3_backend_args = dict(
+            access_key=key,
+            secret_key=secret,
+            **settings.BUNDLE_ASSET_STORAGE_SETTINGS['STORAGE_KWARGS']
         )
 
+        self.s3_backend = S3Boto3Storage(**s3_backend_args)
+
```
pomegranited, line 104: @symbolist  or @kdmccormick -- can you help me understand why the `AssetStorage` class uses the `url_backend` for the url generation, but then uses `default_storage` for the other storage calls? What is the use case for having the storage backend which generates the URLs to be different from the one that actually saves/deletes the files?

I'm trying to figure out why I have to set both `BUNDLE_ASSET_URL_STORAGE_KEY` + `AWS_ACCESS_KEY_ID` + all of the other storage variables, and also `DEFAULT_FILE_STORAGE: storages.backends.s3boto3.S3Boto3Storage` in order to get the blockstore to save its media files to s3. It seems like we only really need the `BUNDLE_ASSET_*` settings (which could default to the `AWS_*` ones if unset).

symbolist: This change was introduced in this PR: https://github.com/openedx/blockstore/pull/101. I hope the description is helpful?

symbolist (commented):
Code changes look fantastic! Will do a round of testing with the LX devstack next.

blockstore/apps/bundles/storage.py
pomegranited: It is, thank you @symbolist !

But in order to run blockstore as an app inside edx-platform, we're going to have to remove the usage of `default_storage` as configured by the `DEFAULT_FILE_STORAGE` and `AWS_*` access variables. I'll work on updating the Blockstore IDA's settings.production to use these app-level defaults, and see what else needs to be changed here.

symbolist:
Everything tested up till now has been working. 👍🏽  Going to do some more testing with complex scenarios, run edx-platform tests etc next.

blockstore/apps/bundles/storage.py
symbolist: I see. 🤔

pomegranited: @symbolist  see 02fa248 for my fix for this issue. I'm still deploying it to the sandbox for testing, but would appreciate your input.

symbolist: The changes make sense to me, thanks.

blockstore/apps/bundles/links.py
```diff
@@ -11,12 +11,18 @@
 import attr
 
 
+def convert_to_uuid(value):
+    if not isinstance(value, UUID):
+        return UUID(value)
+    return value
```
gabor-boros, line 17: Also, this is a not, but how about avoiding the negation here?

```suggestion
    if isinstance(value, UUID):
        return value
    return UUID(value)
```

blockstore/apps/api/tests/test_blockstore_api.py
```diff
@@ -137,12 +194,63 @@ def test_drafts_and_files(self):
...
+            api.get_bundle_file_metadata(bundle.uuid, "nonexistent.txt")
+
+    def test_bundle_version(self):
+        """
+        Test creating, reading, and writing bundle versions and files
+        """
+        coll = api.create_collection("Test Collection")
+        bundle = api.create_bundle(
+            coll.uuid,
+            title="Water 💧 Bundle",
```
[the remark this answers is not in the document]
gabor-boros: So this is for testing the encoding in discussed as part of DB Collation settings, right?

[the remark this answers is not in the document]
pomegranited: @gabor-boros  Yep! This is testing our support for UTF8 characters in the title strings.

blockstore/apps/bundles/links.py
pomegranited: Good suggestion @gabor-boros , done in [054ab81](https://github.com/openedx/blockstore/pull/149/commits/054ab812bfcbcb3613b4786d1a70b37fd3d56414).

gabor-boros:
@pomegranited Thank you for addressing the comments! From my side, this looks good (after a review from @symbolist).

blockstore/apps/api/methods.py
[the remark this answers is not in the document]
pomegranited: Yep -- I can view images added inside HtmlBlocks with this change.

pomegranited:
Thank you for your review @symbolist !  @bradenmacdonald  is going to have a look as core committer, and then I can squash the changes in preparation for merge.

> Once this has been merged, I am assuming we will deploy this to the blockstore service as well? Before merging https://github.com/openedx/edx-platform/pull/29779 i.e..

Hi @jristau1984 I see you're the Reporter for [TNL-8746](https://openedx.atlassian.net/browse/TNL-8746). Can you help us find someone to shepherd this through to release on to edX's blockstore service?

jristau1984:
@pomegranited our team planned our sprint yesterday, so this was not included.  I will ensure this is discussed in our next sprint planning (March 14).  Please let me know if that does not work for your team.  Thanks!

pomegranited:
Brilliant, thank you @jristau1984 ! I think MIT are awaiting this change too, so I'll forward your timeline details to them.

bradenmacdonald (approved):
I read through the code and it looks good to me, thanks! I tested the sandbox on https://github.com/openedx/edx-platform/pull/29779 but will generally rely on @symbolist 's testing here.

pdpinch:
We (at MIT) are eager to see this merged, but our deployment is on maple so we need some additional work before we can actually test it. However, we don't have an existing deployment of blockstore, so most of the steps after "merge and tag" are irrelevant to us. 

We'll keep an eye on this, and https://github.com/openedx/edx-platform/pull/29779 to see when they merge.

jristau1984:
[Update] T&L has pulled this into our sprint starting today.  @doctoryes will be driving the work from our side.  Thanks!

doctoryes:
@symbolist Yes, I will do so. I'm currently working on the infrastructure changes needed for LMS to access the blockstore DB. Once I have an idea of when I'll finish that work, I'll let you know.

symbolist:
@doctoryes thanks!

doctoryes:
@pomegranited : Some additional work has presented itself for this repo's transition to Django application: 

The `ATOMIC_REQUESTS: True` config key was removed from the blockstore DB connection when moving the connection info to edxapp's `DATABASES` Django setting. As edxapp will service many views which do not connect to the blockstore DB, it adds too much unnecessary overhead to create a DB transaction on the blockstore DB upon each Django view serviced by edxapp. Instead, `@transaction.atomic(using=BLOCKSTORE_DB_ALIAS)` decorators need to be added in the blockstore to views which need an atomic DB transaction.

@symbolist is aware of the need for this addition - but just wanted to mention it here as well.

pdpinch:
Is there a new etm (estimate time to merge) for this PR and it's companion https://github.com/openedx/edx-platform/pull/29779?

doctoryes:
@pomegranited @symbolist : The work was merged to allow access from the [edx.org](http://edx.org/) stage and production environments to the blockstore DB. The blockstore DB should now be accessible from Django in both the edxapp web and worker instances using a DB alias of blockstore.

edx-webhook:
@pomegranited 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/blockstore&entry.867055334=https://github.com/openedx/blockstore/pull/149&entry.1484655318=https://github.com/pomegranited&entry.752974735=2022-01-13+05:27&entry.1917517419=2022-03-31+12:43&entry.2133058324=Yes) so we can improve your experience in the future.

symbolist:
@doctoryes Great, thanks!

Before we install this in edxapp, we should deploy the changes to the blockstore service. Can we do this on Monday?

The steps after that should be:
* Add `@transaction.atomic(using=BLOCKSTORE_DB_ALIAS)` decorators.
* Update blockstore package version in https://github.com/openedx/edx-platform/pull/29779 and merge it.
* Turn on the flag on stage and do a round of testing.
* Schedule a window to turn on the flag on production. We will need FFs available to monitor and turn off the flag if needed.

Does TNL have capacity over the next two weeks to support this deployment? CC @pdpinch

pdpinch:
@symbolist In this repo, it can be in the master branch. I don't think BTR is ready to support blockstore as part of nutmeg.

It would be best if the https://github.com/openedx/edx-platform/pull/29779 could be in nutmeg (RC to be cut April 9)

pomegranited:
Thank you for merging this @symbolist ! Could you (or @doctoryes ) tag the latest master so I can reference it in https://github.com/openedx/edx-platform/pull/29779? The [current __version__ is `1.2.0`](https://github.com/openedx/blockstore/blob/master/blockstore/__init__.py#L5), though the existing tags are just for the named releases, so I'm not sure what the format should be.

doctoryes:
@pomegranited I have created the `1.2.0` tag in the blockstore repo, so the edx-platform requirement line can now be:
`-e git+https://github.com/open-craft/blockstore.git@1.2.0#egg=blockstore==1.2.0`
However, ideally this repo would now be built as an installable package and uploaded to PyPI. Is that work something that's planned?

pomegranited:
Brilliant, thank you @doctoryes ! I've updated the blockstore requirement on https://github.com/openedx/edx-platform/pull/29779.

> However, ideally this repo would now be built as an installable package and uploaded to PyPI. Is that work something that's planned?

I agree, but I haven't seen that mentioned anywhere. @farhaanbukhsh or @kaustavb12 is that something that would fit into the BD-14 epic?

doctoryes:
@pomegranited This PR caused the latest Blockstore deploy as an IDA (not as a Django app in edx-platform yet) to fail with this error:
```
  File "/edx/app/blockstore/blockstore/blockstore/apps/bundles/models.py", line 60, in <module>
    from .store import DraftRepo, SnapshotRepo, bytes_from_hex_str
  File "/edx/app/blockstore/blockstore/blockstore/apps/bundles/store.py", line 28, in <module>
    from .storage import default_asset_storage
  File "/edx/app/blockstore/blockstore/blockstore/apps/bundles/storage.py", line 188, in <module>
    default_asset_storage = AssetStorage()
  File "/edx/app/blockstore/blockstore/blockstore/apps/bundles/storage.py", line 141, in __init__
    self.url_backend = LongLivedSignedUrlStorage()
  File "/edx/app/blockstore/blockstore/blockstore/apps/bundles/storage.py", line 99, in __init__
    **settings.BUNDLE_ASSET_STORAGE_SETTINGS['STORAGE_KWARGS']
KeyError: 'STORAGE_KWARGS'
```
I see that `BUNDLE_ASSET_URL_STORAGE_KEY / BUNDLE_ASSET_URL_STORAGE_SECRET` are defined in Django settings - but `BUNDLE_ASSET_STORAGE_SETTINGS` hasn't been added yet.
Questions:
- The code currently expects a new Django setting as shown in [this test code](https://github.com/openedx/blockstore/blob/master/blockstore/apps/bundles/tests/test_storage.py#L44-L52). Was that the intent?
- If so, what is each of those values expected to be? On stage/prod? Is the access_key/secret_key the same as the values in the existing key/secret? What about the other values?

This missing config will need to be fixed first before proceeding with deploying Blockstore-as-a-Django-app in edx-platform. FYI @symbolist

pomegranited:
@doctoryes  Oh bummer.. I missed that issue with my testing :(

I've submitted https://github.com/openedx/blockstore/pull/166 to resolve this issue, but an interim option would be to add the following to edx-internal:

```yaml
BUNDLE_ASSET_STORAGE_SETTINGS:
   STORAGE_CLASS: 'storages.backends.s3boto3.S3Boto3Storage'
   STORAGE_KWARGS: {}
```

As noted in https://github.com/openedx/blockstore/pull/166, the storage kwargs will be pulled from the top-level settings; the only issue here is that I didn't allow that dict to be omitted.

pomegranited:
> Some additional work has presented itself for this repo's transition to Django application:
> 
> The `ATOMIC_REQUESTS: True` config key was removed from the blockstore DB connection when moving the connection info to edxapp's `DATABASES` Django setting. As edxapp will service many views which do not connect to the blockstore DB, it adds too much unnecessary overhead to create a DB transaction on the blockstore DB upon each Django view serviced by edxapp. Instead, `@transaction.atomic(using=BLOCKSTORE_DB_ALIAS)` decorators need to be added in the blockstore to views which need an atomic DB transaction.

@symbolist there's a problem -- if we add `@transaction.atomic(using=BLOCKSTORE_DB_ALIAS)` , then this works for the edxapp deployment which uses a separate `blockstore` database. But anyone running the default deployment will be using the edxapp `default` database, not `blockstore`.

So we're going to have to detect _which_ database is the actual one being used to applying with this wrapper. Ugh!

pomegranited:
NVM @symbolist , you addressed this issue [on Slack](https://openedx.slack.com/archives/C02QN50TYMD/p1651251016902779?thread_ts=1650530641.419139&cid=C02QN50TYMD). Thank you!

PR #183, opened 2022-06-15
[BD-14] fix: remove refrence to access key in logging

connorhaugh:
## Description

So the addition of logging is great, but we are not setting the access key in edx-platform build since we use IAM roles and not key, and therefore the getattr call is erroring. We removed any reference to the access key. 

## Test Instructions

TODO: Include detailed instructions that explain how your reviewers can manually test this code.

## TODOs
- [ ] Squash before merging

### DIFF
--- a/blockstore/apps/bundles/storage.py
+++ b/blockstore/apps/bundles/storage.py
@@ -199,10 +199,9 @@ class AssetStorage(Storage):
 
     def __repr__(self):
         """Used to log details about the configured storage backend"""
-        return 'asset_backend={}(bucket_name={}, access_key={}), url_backend={})'.format(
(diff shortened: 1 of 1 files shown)


bszabo:
I don't know what Jillian's intent was, but suspect this is a belts-and-suspenders strategy, intended to provide maximum value from the next integration exercise. But, given that this attempting to log access keys is breaking the build, the approach is backfiring, and reducing the potential value of the exercise to zero. I'll approve the PR, and let's see if we can get a build deployed to stage in time for our 3:30 exercise.

pomegranited:
Hi @connorhaugh CC @bszabo I've fixed the tests with https://github.com/openedx/blockstore/pull/184 and removed another reference to `access_key` from the logs.

Could you approve, merge, and tag so we can try this one in stage?

PR #203, opened 2022-09-19
fix: github workflow -branch to actual default branch

sarina:
See https://github.com/openedx/docs.openedx.org/pull/135 for more details


bradenmacdonald (approved):
I can confirm that the workflow is currently broken: https://github.com/openedx/blockstore/actions/runs/3124263730/jobs/5076009529#step:3:59

And this looks like the right fix. Thanks!

Merging :)

PR #216, opened 2022-10-27
Fix package name

aht007:
## Description
The name `blockstore` is already taken on PyPI and hence the release failed. Updating the package name in this PR to resolve the issue

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ README = open(os.path.join(os.path.dirname(__file__), 'README.rst')).read()
 CHANGELOG = open(os.path.join(os.path.dirname(__file__), 'CHANGELOG.rst')).read()
 
 setup(
-    name='blockstore',
+    name='openedX-blockstore',
     version=VERSION,
     description="""Blockstore is a storage system for learning content in Open edX.""",
     long_description=README + '\n\n' + CHANGELOG,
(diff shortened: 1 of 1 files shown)


aht007:
As per the conversation with @nedbat the prefix should be openedx as the repo is hosted in openedX GH organization.
https://edx-internal.slack.com/archives/CG7FM3BLY/p1666881497813729?thread_ts=1666861508.868599&cid=CG7FM3BLY

UsamaSadiq (approved):

connorhaugh:
A quick question: will this be a breaking change in places where we install blockstore (edx-platform)? How are we effectively making that transition?

nedbat:
blockstore is currently installed from GitHub using a pin:
https://github.com/openedx/edx-platform/blob/master/requirements/edx/github.in#L74
```
-e git+https://github.com/openedx/blockstore.git@1.2.1#egg=blockstore==1.2.1  # Note: Blockstore 1.2.2 is failing.
```
so this will not break edx-platform.

aht007:
@connorhaugh this will not break anything as we are just changing the name of PyPi distribution. The imports will still remain the same by this change. As Ned also mentioned that its installed using a pin in edx-platform so if any breaking change still makes it to the release it wouldn't affect edx-platform. Also I have followup ticket to install blockstore from PyPI in edx-platform so I will be testing the changes in the scope of that ticket as well.
Just waiting for the review now so that I can go ahead and merge and release it to PyPI.

connorhaugh (approved):

PR #235, opened 2023-02-06
fix: allow bundles to be created via django admin

kdmccormick:
## Description

fix: allow bundles to be created via django admin

Previously, trying to create a bundle would lead to the
message:

    IntegrityError at /admin/bundles/bundle/add/
    (1048, "Column 'collection_id' cannot be null")

because `collection` was indeed a required field, but
it was marked as editable=False, so it one could not
actually set it when making a new bundle.

With this change, we set editable=True, and add
some logic to the admin site so that the
`collection` field can be set when a bundle
is first created, but not afterwards.

This will trigger a migration, but since `editable`
is purely a Django-level validation field, no
changes should be seen in the MySQL schema.

N/A

### Setup

Get blockstore running in a dev env, and make sure at least one Collection is created. I chose to use Tutor plus the [library authoring Tutor plugin](https://github.com/openedx/frontend-app-library-authoring/tree/master/tutor-contrib-library-authoring-mfe), which uses Blockstore as a CMS app.

### Reproducing the bug

In CMS admin, go to add a bundle (http://studio.local.overhang.io:8001/admin/bundles/bundle/add/)

Clicking "Save" from here:

image

Gives:
```
IntegrityError at /admin/bundles/bundle/add/

(1048, "Column 'collection_id' cannot be null")
```
### Testing the fix

In CMS admin, go to add a bundle (http://studio.local.overhang.io:8001/admin/bundles/bundle/add/)

Notice that you are now able to select a Collection. Clicking Save should work:

image
image

Clicking into the Bundle you created should display a link to the Collection, but not allow you to edit it:

image

## TODOs

Tag 1.3.0 once this merges.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,17 +16,21 @@ Unreleased
 
 *
 
-[1.1.0] - 2021-10-25
+[1.3.0] - 2023-02-06
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Added
 _____
 
-* Move apps/api to apps/rest_api.
-* Copy edx-platform/openedx/core/lib/blockstore_api to blockstore/apps/api.
-  This code has been copied over so it is easier to review the Python API
-  implementation in development.
-* Make code into an installable package.
+* Adds support for installing this package as a wheel (``pip install openedx-blockstore``) rather than having to use editable mode (``pip install -e ./blockstore``).
+
+Changed
+_______
+
+* Fixes a bug where a new bundle's collection could not be set via Django admin.
+* Various configuration and build tooling fixes. See commit log of 1.2.0...1.3.0 for full details.
+
+
 
 [1.2.0] - 2021-01-25
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


kdmccormick:
Hey @pomegranited , could someone over at OpenCraft take give this a review ? Not high-urgency.

pomegranited:
Sure thing @kdmccormick , thanks for the ping :)

@bradenmacdonald do you have time to review this?

bradenmacdonald:
@kdmccormick @pomegranited Yes I'll be happy to review this. Sorry for the delay though, I'm currently digging myself out of a hole with my TODO list.

bradenmacdonald (approved):
Thanks so much @kdmccormick. Please merge at your convenience or let me know if you need me to.

:+1:

- [x] I tested this: in the CMS django admin as described.
- [x] I read through the code
- [x] I checked for accessibility issues
- [x] Includes documentation: changelog etc.


The only issue I ran into is I somehow had two instances of blockstore installed at the same time, one with the egg/name `openedx_blockstore` and one with `blockstore`. I kept getting the old version until I manually removed it, then this one became active. But I assume that is a temporary issue that will soon be resolved once we're consistently installing from PyPI.

kdmccormick:
Thanks Jill and Braden.

> The only issue I ran into is I somehow had two instances of blockstore installed at the same time, one with the egg/name openedx_blockstore and one with blockstore. I kept getting the old version until I manually removed it, then this one became active. But I assume that is a temporary issue that will soon be resolved once we're consistently installing from PyPI.

Yeah, that sounds transitory to me, but let me know if you keep seeing it in the future!

awais786:
@kdmccormick you might need a new migration for this PR. migration test fail in this platform.
```
FAILED common/djangoapps/util/tests/test_db.py::MigrationTests::test_migrations_are_in_sync - 
assert 'No changes detected' in "Migrations for 'bundles':\n  
/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/blockstore/apps/bundles/migrati....deletion.CASCADE, related_name='bundles', related_query_name='bundle', to='bundles.collection'),\n        ),\n    ]\n"
```

kdmccormick:
Whoops, thanks for catching that @awais786 . I just put up a fix: https://github.com/openedx/blockstore/pull/241

PR #241, opened 2023-03-06
fix: add missing migration

kdmccormick:
Migration should have been added in
https://github.com/openedx/blockstore/pull/235

The change is purely a Django-level validation
field. No changes should be seen in the MySQL schema.


bradenmacdonald (approved):
Thanks for fixing it @kdmccormick

PR #256, opened 2023-05-25
docs: add local plugin dev to README

connorhaugh:
This adds documentation for how users can develop with blockstore locally in devstack as a python plugin.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -133,6 +133,19 @@ If you want to develop blockstore as a local python package installed in edx-pla
       
 Seeing your changes will sometimes require running make requirements and then restarting the container.
 
+You're probably also going to want to create a collection for using content libraries, using:
+
+   .. code::
+   
+       # Create a "Collection" that new content libraries / xblocks can be created within:
+        docker exec -t edx.devstack.blockstore bash -c "source ~/.bashrc && echo \"from blockstore.apps.bundles.models import Collection; coll, _ = Collection.objects.get_or_create(title='Devstack Content Collection', uuid='11111111-2111-4111-8111-111111111111')\" | ./manage.py shell"
+       # Create an "Organization":
+       docker exec -t edx.devstack.lms bash -c "source /edx/app/edxapp/edxapp_env && echo \"from organizations.models import Organization; Organization.objects.get_or_create(short_name='DeveloperInc', defaults={'name': 'DeveloperInc', 'active': True})\" | python /edx/app/edxapp/edx-platform/manage.py lms shell"
+       
+The Library Authoring MFE will want to know about that collection id `using a ENV variable. <https://github.com/openedx/frontend-app-library-authoring/blob/d590aa2eb54c94b39d94f2ba12a6b458082c2e5e/.env#L19>`_
+       
+Then restart Studio and the LMS (``make dev.restart-devserver.lms dev.restart-devserver.studio``).
+
 -------------------------------------------------------
 Running and testing as a separate service
 -------------------------------------------------------
--- a/README.rst
+++ b/README.rst
@@ -97,7 +97,7 @@ If you want to develop blockstore as a local python package installed in edx-pla
 
 #. Prerequisite: Have an Open edX `Devstack <https://github.com/openedx/devstack>`_ properly installed and working. Your devstack must use the Nutmeg release of Open edX (or newer) or be tracking the ``master`` branch of ``edx-platform``.
 
-#. To run blockstore as an app inside of Open edX, enable it using the waffle switch `blockstore.use_blockstore_app_api <https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-blockstore.use_blockstore_app_api>`_.
+#. To run blockstore as an app inside of Open edX, enable it by activating the waffle switch `blockstore.use_blockstore_app_api in LMS or Studio admin. <https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-blockstore.use_blockstore_app_api>`_.
 
 #. Clone this repo in the ``src/`` directory, a sibling directory of your devstack folder.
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
-Running and testing as a separate service (development)
+Development in Devstack
+-------------------------------------------------------
+
+If you want to develop blockstore as a local python package installed in edx-platform, use the following steps:
+
+#. Prerequisite: Have an Open edX `Devstack <https://github.com/openedx/devstack>`_ properly installed and working. Your devstack must use the Nutmeg release of Open edX (or newer) or be tracking the ``master`` branch of ``edx-platform``.
+
+#. To run blockstore as an app inside of Open edX, enable it using the waffle switch `blockstore.use_blockstore_app_api <https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-blockstore.use_blockstore_app_api>`_.
+
```
bszabo, line 101: I'm assuming it doesn't matter whether this is done in LMS admin or Studio admin, but might be best to be clear on the point.

bszabo: The admin waffle switch form allows you to activate a switch, not enable it. Might be best to use 'activate' as the verb here.

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+#. Clone this repo in the ``src/`` directory, a sibling directory of your devstack folder.
+
+#. In your devstack directory, run the following commands in order:
+
+   .. code::
+
```
bszabo, line 107: Is there an assumption about virtual environments here? If using a virtual environment, should be activating it before these operations.

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+   .. code::
+      
+      # bring up the lms +studio containers
+      make dev.up.large-and-slow
+      # Bring up the studio shell
+      make dev.shell.studio
+      # Unistall the pypi blockstore
+      pip uninstall openedx-blockstore #You will need to confirm the unistall with "Y".
+      #Install your local blockstore.
```
bszabo, line 114: echoes "  Successfully uninstalled openedx-blockstore-1.3.1" (or some such) on success

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+      pip install -e /edx/src/blockstore/  
+      exit
```
bszabo, line 116: echoes "Successfully installed openedx-blockstore-1.3.1" (or some such) on success

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+    
+      #restart the containers.
+      make dev.restart-devserver.lms dev.restart-devserver.studio
+
+      # you can see that your env now installed blockstore from a local directory by running:
+      make dev.shell.lms
```
bszabo, line 130: Clearer to stat that `openedx-blockstore` is now installed from a local directory, as that's the key used in the list

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+      pip list
+       
+      
+Seeing your changes will sometimes require running make requirements and then restarting the container.
+
+You're probably also going to want to create a collection for using content libraries, using:
+
```
bszabo, line 137: What is your shell reference point now? The previous set of instructions were inside a container shell. Are we now back to the devstack shell?

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+   .. code::
```
bszabo, line 138: What are your assumptions about the existence and state of a blockstore container prior to this point? What about now?

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+      
+      # do the same in the LMS
+      make dev.shell.lms
+      # Unistall the pypi blockstore
+      pip uninstall openedx-blockstore #You will need to confirm the unistall with "Y".
+      #Install your local blockstore.
+      pip install -e /edx/src/blockstore/  
+      exit
+    
+      #restart the containers.
```
bszabo, line 126: Restart the servers

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+
+You're probably also going to want to create a collection for using content libraries, using:
+
+   .. code::
+   
+       # Create a "Collection" that new content libraries / xblocks can be created within:
+        docker exec -t edx.devstack.blockstore bash -c "source ~/.bashrc && echo \"from blockstore.apps.bundles.models import Collection; coll, _ = Collection.objects.get_or_create(title='Devstack Content Collection', uuid='11111111-2111-4111-8111-111111111111')\" | ./manage.py shell"
+       # Create an "Organization":
+       docker exec -t edx.devstack.lms bash -c "source /edx/app/edxapp/edxapp_env && echo \"from organizations.models import Organization; Organization.objects.get_or_create(short_name='DeveloperInc', defaults={'name': 'DeveloperInc', 'active': True})\" | python /edx/app/edxapp/edx-platform/manage.py lms shell"
+
```
bszabo, line 144: Some deprecation warnings are expected

README.rst
```diff
@@ -90,7 +90,64 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
...
+   
+       # Create a "Collection" that new content libraries / xblocks can be created within:
+        docker exec -t edx.devstack.blockstore bash -c "source ~/.bashrc && echo \"from blockstore.apps.bundles.models import Collection; coll, _ = Collection.objects.get_or_create(title='Devstack Content Collection', uuid='11111111-2111-4111-8111-111111111111')\" | ./manage.py shell"
+       # Create an "Organization":
+       docker exec -t edx.devstack.lms bash -c "source /edx/app/edxapp/edxapp_env && echo \"from organizations.models import Organization; Organization.objects.get_or_create(short_name='DeveloperInc', defaults={'name': 'DeveloperInc', 'active': True})\" | python /edx/app/edxapp/edx-platform/manage.py lms shell"
+       
+The Library Authoring MFE will want to know about that collection id `using a ENV variable. <https://github.com/openedx/frontend-app-library-authoring/blob/d590aa2eb54c94b39d94f2ba12a6b458082c2e5e/.env#L19>`_
+       
+Then restart Studio and the LMS (``make dev.restart-devserver.lms dev.restart-devserver.studio``).
+
```
bszabo, line 148: The testing part of the readme is still a bit thin. What steps to follow and what to look for to confirm that a) the IDA is no longer running; b) the Django app is; and c) the Django app blockstore is running as expected.

bszabo:
There's a lint commit error that needs attention

README.rst
connorhaugh: You actually totally could do this without a venv. I know this because I accidentally did.

connorhaugh: This is such a base assumption for working in openedx I don't feel like I need to include it.

README.rst
connorhaugh: I think adding that level of hand-holding might just lead to confusion (what if we remove the deprecation warnings, and then people don't see them).

README.rst
connorhaugh: This guide doesn't care that you ever ran blockstore as a separate service. You could be running the blockstore service at the same time, but the waffle flag makes it such that it doesn't matter. This guide replaces that setup for local dev. 
I added a part to direct the user to set up content libraries v2. That is how blockstore is used and the system's effectiveness can be tested.

README.rst
connorhaugh: > There's a lint commit error that needs attention

I'll just squash and merge.

connorhaugh:
> There's a lint commit error that needs attention
I'll just squash when I merge.

README.rst
bszabo: The link's text inappropriately includes the phrase "in LMS or Studio Admin", making it hard to see that the clarification isn't just part of the switch name.

README.rst
bszabo: Picked up an extra "is" that makes the sentence not parse right. Better as  "... that your env now installs openedx-blockstore from the local directory ..."

README.rst
bszabo: I don't think it's superfluous to indicate that with Blockstore running this way a `docker ps` should not show any running blockstore container. While that instruction may be obvious to many, it will help some.

bszabo (approved):

PR #286, opened 2023-09-06
docs: add "create collection" to usage docs

connorhaugh:
## Description

TODO: Include a detailed description of the changes in the body of the PR

## Test Instructions

TODO: Include detailed instructions that explain how your reviewers can manually test this code.

## TODOs
- [ ] Squash before merging

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,6 +89,13 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the 
 
 If you need to run blockstore as a separate service (e.g. for scalability or performance reasons), you can deploy blockstore in production using `the blockstore ansible role <https://github.com/openedx/configuration/tree/master/playbooks/roles/blockstore>`_.
 
+you'll want to also create a collection and organization.
+
+# Create a "Collection" that new content libraries / xblocks can be created within:
+ docker exec -t edx.devstack.lms bash -c "source ~/.bashrc && echo \"from blockstore.apps.bundles.models import Collection; coll, _ = Collection.objects.get_or_create(title='Devstack Content Collection', uuid='11111111-2111-4111-8111-111111111111')\" | ./manage.py lms shell"
+# Create an "Organization":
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -89,6 +89,13 @@ By default, blockstore is run as an app inside of Open edX. Enable it using the
 
 If you need to run blockstore as a separate service (e.g. for scalability or performance reasons), you can deploy blockstore in production using `the blockstore ansible role <https://github.com/openedx/configuration/tree/master/playbooks/roles/blockstore>`_.
 
+you'll want to also create a collection and organization.
+
+# Create a "Collection" that new content libraries / xblocks can be created within:
+ docker exec -t edx.devstack.lms bash -c "source ~/.bashrc && echo \"from blockstore.apps.bundles.models import Collection; coll, _ = Collection.objects.get_or_create(title='Devstack Content Collection', uuid='11111111-2111-4111-8111-111111111111')\" | ./manage.py lms shell"
+# Create an "Organization":
+docker exec -t edx.devstack.lms bash -c "source /edx/app/edxapp/edxapp_env && echo \"from organizations.models import Organization; Organization.objects.get_or_create(short_name='DeveloperInc', defaults={'name': 'DeveloperInc', 'active': True})\" | python /edx/app/edxapp/edx-platform/manage.py lms shell"
+
```
rayzhou-bit, line 98: Is the following from a later section required as well?

`
The Library Authoring MFE will want to know about that collection id `using a ENV variable. <https://github.com/openedx/frontend-app-library-authoring/blob/d590aa2eb54c94b39d94f2ba12a6b458082c2e5e/.env#L19>`_
       
Then restart Studio and the LMS (``make dev.restart-devserver.lms dev.restart-devserver.studio``).
`

rayzhou-bit (changes requested):
I think you should remove Collection/Organization setup in the later section to keep the doc clear.

bradenmacdonald:
I've said this before, but I think it's worth repeating: The "Collection" concept is not really used for anything, and doesn't involve many lines of code. We should take the time now to just delete the Collection model entirely instead of continuing to maintain code that isn't useful and documenting how to work around it.

